PR #11, opened 2018-05-18
`run` command

mih:


datalad_container/containers_run.py
```diff
@@ -0,0 +1,112 @@
...
+@build_doc
+# all commands must be derived from Interface
+class ContainersRun(Interface):
+    # first docstring line is used a short description in the cmdline help
+    # the rest is put in the verbose help and manpage
+    """Drop-in replacement of 'run' to perform containerized command execution
+
+    Container(s) need to be configured beforehand (see containers-add).
+    If only one container is known, it will be selected automatically,
+    otherwise a specific container has to be selected.
```
kyleam: minor, but I think replacing the second "selected" with "specified" would be clearer.

datalad_container/containers_run.py
```diff
@@ -0,0 +1,112 @@
...
+
+_run_params = dict(
+    Run._params_,
+    container_name=Parameter(
+        args=('-n', '--container-name',),
+        metavar="NAME",
+        doc="""Specify the name of a known container to use for execution,
+        in case multiple containers are configured."""),
+    # have to override the cmd parameter, because in `run` it says
+    # SHELL COMMAND but with a container it could be something totally
```
kyleam: I don't know that the SHELL in run's `metavar='SHELL COMMAND'` really adds that much.  Perhaps we should just change it to `metavar='COMMAND'` there and make sure Run's class docstring mentions explicitly that it's a shell command.

mih: That would be a solution.

datalad_container/containers_run.py
mih: Done

PR #12, opened 2018-05-18
Implement a `remove` command

mih:
The go through the config and wipe out references to a particular container. And maybe also `git rm` the container image, if it is in the same dataset.

### DIFF
--- a/datalad_container/containers_remove.py
+++ b/datalad_container/containers_remove.py
@@ -1,0 +1,99 @@
+
+__docformat__ = 'restructuredtext'
+
+import logging
+import os.path as op
+
+from datalad.interface.base import Interface
+from datalad.interface.base import build_doc
+from datalad.support.param import Parameter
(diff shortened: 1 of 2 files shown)


datalad_container/containers_remove.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # parameters of the command, must be exhaustive
+    _params_ = dict(
+        dataset=Parameter(
+            args=("-d", "--dataset"),
+            doc="""specify the dataset to query. If no dataset is given, an
+            attempt is made to identify the dataset based on the current
+            working directory""",
+            constraints=EnsureDataset() | EnsureNone()),
+        label=Parameter(
```
kyleam: Shouldn't this be "name" for consistency with add/run?

mih: Yes, but I actually want to swap that in `add` too. So consistency will be restored.

PR #21, opened 2018-05-19
BF: Replace a leftover 'name' from the name->label conversion

kyleam:


mih:
Ah, crossed in cyberspace.. There were a few more. Coming in #23

PR #25, opened 2018-05-19
BF: run: Use the relative path of the container

kyleam:
The image path in the command and the path given to inputs shouldn't
be an absolute path because that won't work across different clones.


kyleam:
Updated for merge of datalad/datalad#2509.

PR #29, opened 2018-05-20
BF: Propagate run's failures

kyleam:
Without on_failure='ignore', containers-run will fail silently when
the run command underneath fails.

---

```
% singularity pull -n bbox.img docker://busybox:latest
% datalad add bbox.img
% datalad containers-add bbox -i bbox.img --call-fmt singularity exec {img} {cmd}
% touch dirt
```

Before change

```
% datalad containers-run --container-label bbox touch out01
<nothing>
```

After change

```
% datalad containers-run --container-label bbox touch out01
run(impossible): /home/kyle/scratch/dl-docker (dataset) [unsaved modifications present, cannot detect changes by command]
```

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -100,5 +100,6 @@ class ContainersRun(Interface):
                 inputs=inputs,
                 outputs=outputs,
                 message=message,
-                expand=expand):
+                expand=expand,
+                on_failure="ignore"):
             yield r


yarikoptic:
Shouldn't then something be fixed in datalad itself... "Fail" if "ignore" but be silent otherwise makes little sense to me

kyleam:
> Shouldn't then something be fixed in datalad itself... "Fail" if "ignore" but be silent otherwise makes little sense to me

Perhaps, but I'm not sure what that would be.  IIUC things are working as designed.  `ContainersRun` is pulling `Run`s results through, and it is responsible for rendering an `IncompleteResultsError`:

https://github.com/datalad/datalad/blob/e44c3cb3f1d461e5d103758badd2869998a7ce4c/datalad/cmdline/main.py#L513-L523

But it can't do that if `Run.__call__` raises an exception, so it has to set `on_failure="ignore"`.

yarikoptic:
but shouldn't at least some error message should be "rendered" by some default renderer if an uncaught exception was raised?

PR #32, opened 2018-05-21
ENH: add: Support call formats that have flags

kyleam:
Without REMAINDER, --call-fmt values that have flags (e.g.,
`singularity exec -e {img} {cmd}`) fail because argparse tries to
parse the flag.

This keeps the --call-fmt flag rather than making it a positional
argument.  The advantage of using a flag is that it's more explicit
and that LABEL isn't required to come immediately before the format
arguments.  Also, datalad's parser currently fails on positional
arguments with dashes (datalad/datalad#2269), so we'd need to rename
the variable.

Re: https://github.com/datalad/datalad-container/pull/30#issuecomment-390626689

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -6,6 +6,7 @@ import re
 import logging
 import os.path as op
 from simplejson import dumps
+from argparse import REMAINDER
 
 from datalad.interface.base import Interface
 from datalad.interface.base import build_doc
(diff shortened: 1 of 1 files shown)


mih:
@kyleam LGTM.

Test failure is due to https://github.com/datalad/datalad/issues/2513 -- which is a complete mystery.

PR #35, opened 2018-05-28
BF: Support '{pwd}' macro and use it to bind mount pwd for singularity

mih:
There is a problem: If we use absolute paths for expansion the resulting run record will not be reproducible.  But if we use a relative paths, singularity will say

```
WARNING: Not mounting requested bind point (already mounted in container): .
```

and not do anything. But eventually fail in some scenarios. Example:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/data/psyinf/scratch/multires3t/bids/sourcedata/REDACTED/dicoms'
```

(same command with abspath(pwd) works just fine).

```
singularity exec .datalad/environments/conversion/image mount
singularity exec --bind . .datalad/environments/conversion/image mount
singularity exec --bind $(pwd) .datalad/environments/conversion/image mount
```

First two yield identical results, diff to the last one is:

```diff
--- plain       2018-05-28 13:08:39.640941047 +0200
+++ abs 2018-05-28 13:09:13.744305938 +0200
@@ -11,6 +11,7 @@
 <snipp>
+zippy/scratch/data/psyinf on /home/data/psyinf/scratch/multires3t/bids type zfs (rw,nosuid,nodev,noatime,xattr,noacl)
 <snipp>
```

Leading to:

```
% singularity exec .datalad/environments/conversion/image ls $(pwd)
WARNING: Not mounting requested bind point (already mounted in container): .
ls: cannot access '/home/data/psyinf/scratch/multires3t/bids': No such file or directory
```

vs.

```
% singularity exec --bind $(pwd) .datalad/environments/conversion/image ls $(pwd)
<snip expected list of directory content>
```

Any ideas on how to solve that? @kyleam

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -87,6 +87,8 @@ class ContainersRun(Interface):
                     fullcmd.extend(cmd)
                 elif c == '{img}':
                     fullcmd.append(image_path)
+                elif c == '{pwd}':
+                    fullcmd.append(pwd)
                 else:
                     fullcmd.append(c)
             cmd = fullcmd
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -58,7 +58,11 @@ def _guess_call_fmt(ds, name, url):
     if url is None:
         return None
     elif url.startswith('shub://'):
-        return ['singularity', 'exec', '{img}', '{cmd}']
+        # use bind to always have the command pwd mounted
+        # this will typically be the dataset root
+        # if the dataset is not in the home drive, execution would
+        # fail otherwise
+        return ['singularity', 'exec', '--bind', '{pwd}', '{img}', '{cmd}']
 
 
 @build_doc


kyleam:
> There is a problem: If we use absolute paths for expansion the resulting run record will not be reproducible. But if we use a relative paths, singularity will say
>
> WARNING: Not mounting requested bind point (already mounted in container): .
>
> [...s]
>
> % singularity exec --bind $(pwd) .datalad/environments/conversion/image ls $(pwd)
> <snip expected list of directory content>

An additional complication when mounting within home:

```
% cd /tmp/
% singularity pull -n bbox docker://busybox:latest
% cd ~/scratch
% singularity exec --bind $(pwd) /tmp/bbox pwd
ERROR  : Not mounting requested bind point (already mounted in container): /home/kyle/scratch
ABORT  : Retval = 255
% singularity exec -c --bind $(pwd) /tmp/bbox pwd
ERROR  : Not mounting requested bind point (already mounted in container): /home/kyle/scratch
ABORT  : Retval = 255
% singularity exec /tmp/bbox pwd
/home/kyle/scratch
% singularity exec -c /tmp/bbox pwd
/home/kyle
```

> Any ideas on how to solve that? @kyleam

Two options come to mind:

  * Add a format field to `datalad run` that gets expanded to the absolute pwd, in line with what we're planning for `{input}/{output}`.

  * In the cases where `--bind` is necessary, use a call_fmt that calls a wrapper that will set up the pwd mount at run time.  This is what [the docker adapter does][0] (for both pwd and uid/gid).

The first one feels wrong because it's adding a field to `datalad run` for what might be a container-specific use case.  The second one seems like a more general solution, but it is more cumbersome and perhaps more confusing since it departs from a plain `singularity ...` call.

[0]: https://github.com/datalad/datalad-container/blob/ff991397fda79361b999ba33f3e7e82e07f23a2d/datalad_container/adapters/docker.py#L125-L135

datalad_container/containers_run.py
```diff
@@ -87,6 +87,8 @@ def __call__(cmd, container_name=None, dataset=None,
                     fullcmd.extend(cmd)
                 elif c == '{img}':
                     fullcmd.append(image_path)
+                elif c == '{pwd}':
```
kyleam: This wouldn't work in the common case of `--binding src:dest`.  When thinking about the `{input}/{output}` fields for `datalad run`, I've wondered whether we should switch to just always running a string  command and `shell=True`, creating the string from the list of args if necessary.  We could then always use `str.format` to expand the fields.

mih: Yeah, I am gravitating more and more in this direction too. It would simplify things quite a bit. Can you this?

mih:
re wrapper vs {pwd} placeholder:

I tend to prefer the placeholder solution.It is not just containers, it is any command that needs to eat absolute paths, while we *never* want to record those, if we are aiming for (potential) reproducibility. Having said that, in the case of a container in a dataset the situation is clear. But in the general case, such relative paths might not enable any better reproducibility, because the dataset needs to be in the exact same place, or some resources needs to match this requirement. In such cases, our transformation would merely make this fact less transparent.

I am torn.

I guess an explicit placeholder at least makes it clear that *any* PWD is fine.

datalad_container/containers_run.py
kyleam: > I am gravitating more and more in this direction too. It would simplify things quite a bit. Can you this?

Sure, I'll open a PR for this today.

kyleam:
> I tend to prefer the placeholder solution.It is not just containers, it is any command that needs to eat absolute paths

Fair enough.  It's probably useful enough that run should support it.  I'll plan to add '{pwd}' as a placeholder to `datalad run`.

mih:
Superseded by https://github.com/datalad/datalad/pull/2594

PR #50, opened 2018-08-04
ENH: default call fmt for singularity to include -e -c

yarikoptic:
If we are aiming for reproducibility we should minimize if not eliminate
any influence of our working environment.  Singularity by default propages
environment, bind mounts home etc.  With `-e -c` (re)execution should become
more reproducible

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -56,7 +56,7 @@ def _guess_call_fmt(ds, name, url):
     if url is None:
         return None
     elif url.startswith('shub://'):
-        return 'singularity exec {img} {cmd}'
+        return 'singularity exec -e -c {img} {cmd}'
     elif url.startswith('dhub://'):
         return 'python -m datalad_container.adapters.docker run {img} {cmd}'
 


yarikoptic:
Yeap, will do, or feel welcome to push a fix. Also we need to look into making a unittests which actually tests for having usable default - I guess there is none ATM

yarikoptic:
Forgot about this one
Re `/tmp` - I guess we should not just bind current directory as `/tmp` since then any temporary output leftovers might pollute our pristine `pwd`.  

And we cannot use just some temporary directory under `/tmp` as it wouldn't yet exist within:
```shell
$> singularity exec -c -e -B $PWD:/tmp/workdir --pwd /tmp/workdir ./neurodebian-latest.img ls /tmp 
ERROR  : Could not change directory to: /tmp/workdir
ABORT  : Retval = 255
```
(in real implementation `workdir` would be with some randomness to its name) unless we pre-create it and locally and bind mount the entirety of the `/tmp` (well - local $TMPDIR ideally):
```shell
$> mkdir -p $TMPDIR/workdir; singularity exec -c -e -B $TMPDIR:/tmp --pwd /tmp/workdir ./neurodebian-latest.img pwd         
/tmp/workdir
```
But I wondered, why don't we just bind mount `pwd` into `/home` which should exist within container?! also for all needs and purposes we should provide a bind mount for `/tmp` to point to a clean temporary directory, so something like this

```shell
$> tmpdir=$(mktemp -d); singularity exec -c -e -B "$tmpdir:/tmp" -B "$PWD:/home" --pwd /home ./neurodebian-latest.img pwd; rm -rf "$tmpdir"
```
sounds like a best example recipe.  If agreed, then we probably need to provide some support within `containers-run` for support of `{tmpdir}` within the cmd template?  I think it is cheap enough to always mkdir a temporary directory and `rmtree` it afterwards.  Then the command would be

```
singularity exec -c -e -B {{tmpdir}}:/tmp -B {{pwd}}:/home --pwd /home {img} {cmd}
```

- what do you think @kyleam ?
- don't we need explicit `""` around all the args which could contain spaces or they get autoescaped/quoted? (didn't try in obscure situation, tests should definitely test under tricky TMPDIRs, like we do in DataLad "core")

PS @AdinaWagner - do you have an idea why we haven't tried today to just bindmount your workdir to smth like `/tmp` or `/home`? ;-)  I would blame Friday -- we should officially announce it a part of the weekend

yarikoptic:
quick one:
> I lean towards not making the working directory in "/home". That way, we can leave $HOME untouched and it will still exist (empty when `-c` is used). Instead, I think we could do something like
> 
> ```
> singularity exec -c -e --pwd /datalad-container-dataset -B "{{pwd}}:/datalad-container-dataset" {img} {cmd}
> ```
Shouldn't then `/datalad-container-dataset` already exist in the image, or we would be out of luck?

kyleam:
> Shouldn't then `/datalad-container-dataset` already exist in the image, or we would be out of luck?

My testing indicates that we're fine, but perhaps it's singularity version dependent?

```
% singularity --version
2.5.2-dist
```

yarikoptic:
Or overlay support in kernel- dependent... Sure thing ideally we should look forward but many HPCs might not have it? I will double check on Hydra with oldstable Debian and 3.14 iirc kernel

kyleam:
> Sure thing ideally we should look forward but many HPCs might not have it? I will double check on Hydra with oldstable Debian and 3.14 iirc kernel

In terms of singularity versions, note that versions before 2.4.3 are broken wrt using `-c/-C` and `--pwd` together (I only know this because I filed the bug), so that's probably the floor of what we want to support.

yarikoptic:
yeap...
as for kernel: 
- on hydra
```
head1:/data/movieloc/backup_store/saccs
$> singularity exec --pwd /buga --bind $PWD:/buga .datalad/environments/BIDSSacc/image /bin/bash
WARNING: Non existent bind point (directory) in container: '/apps'
WARNING: Skipping user bind, non existent bind point (directory) in container: '/buga'
ERROR  : Could not change directory to: /buga
ABORT  : Retval = 255
$> lsmod | grep overlay
$> uname -a
Linux head1 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2 (2016-04-08) x86_64 GNU/Linux
```
I will be updating this beast soon (now) so we would gain overlays as well

while on "new" discovery they seems to have overlay even with older kernel:
```shell
*[yhalchen@discovery7 ~]$ singularity exec --pwd /buga --bind $PWD:/buga ./git-annex-dev.img  /bin/bash
yhalchen@discovery7:~$ ls /buga
 -						  7.20181105+git22-g4c7236c58-1~ndall+1~nd+1.tgz   datalad-singularity-outofmem.tgz				       p.patch
...
yhalchen@discovery7:~$ exit
[yhalchen@discovery7 ~]$ uname -a                                                                     
Linux discovery7.hpcc.dartmouth.edu 3.10.0-862.14.4.el7.x86_64 #1 SMP Wed Sep 26 15:12:11 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
[yhalchen@discovery7 ~]$ lsmod | grep overlay                                                       
overlay                71964  1 
```
So overall may be we should aim/assume overlay supports and just indeed bind mount some non-existing in the image dedicated directory... I wish we had more HPC samples to make judgement

yarikoptic:
Note:  if we provide sanitized $HOME (`-H`) then any "datalad" invocation would be doomed since there would be no git configuration we demand... So for an ideal reproducible invocation we would need to pre-craft `$HOME` with our deep ideas or demand user to provide one

yarikoptic:
I increasingly need to be able to execute singularity in a sanitized environment, and typically with some "temporary" home (in my cases I didn't need git config or ssh stuff inside). In those having smth like `{tmpdir}` placeholder (discussed above) would have been sufficient.

semi-sanitized is the solution having a persistent directory (e.g. `.datalad/environments/<name>/home`).  This way user has full control on what to populate it with BUT its content might change, upon each run, and thus we would need to keep it under version control AND may be optionally instruct us to "reset" such a home after each run to not grow side-effects of execution.  Since home directory specification to singularity requires full path, we would need to have something like `{toppath}` to be used within `.datalad/config`

kyleam:
> I increasingly need to be able to execute singularity in a sanitized environment, and typically with some "temporary" home (in my cases I didn't need git config or ssh stuff inside). In those having smth like `{tmpdir}` placeholder (discussed above) would have been sufficient.

`{tmpdir}` seems like a good addition to run's placeholders.  Adding it on containers-run's end wouldn't work because then you'd be hard-coding it in the recorded command.

For anything more complicated, I'd lean towards custom command wrappers (and improving support or them in datalad-container or elsewhere).

yarikoptic:
> For anything more complicated, I'd lean towards custom command wrappers (and improving support or them in datalad-container or elsewhere).

yeah, I can generate a shim (good idea) and have it e.g. as `.datalad/bin/singularity` but I would still need `{toppath}` or alike to point to it within `.datalad/config` to execute from anywhere by top directory

yarikoptic:
FTR: In the [ReproNim/containers](https://github.com/ReproNim/containers) dataset I am experimenting with the wrapper/setup to assist running containerized computation with better "sanitation" - `-e -c` options, custom `$HOME`, dedicated temporary directory to be used for `/tmp` and `/var/tmp` within container.  I guess depending on what I arrive there with, it might be worth generalizing it here.

yarikoptic:
FWIW, here is the "executioner" script https://github.com/ReproNim/containers/blob/master/scripts/singularity_cmd which bind mounts minimalistic HOME directory with rudimentary .gitconfig and .gitignore: https://github.com/ReproNim/containers/tree/master/binds/HOME .  But may be ideally that .gitconfig should be [autogenerated by the executioner script](https://github.com/ReproNim/containers/issues/10)

yarikoptic:
too old, nobody cares, shims like ReproNim/containers are used instead

PR #53, opened 2018-10-23
ENH: Allow to also specify container by path rather than just its name

bpoldrack:


datalad_container/containers_run.py
```diff
@@ -68,10 +68,18 @@ def __call__(cmd, container_name=None, dataset=None,
...
-            # anything else is an error
-            raise ValueError(
-                'Container selection impossible: not specified or unknown '
-                '(known containers are: {})'.format(list(containers.keys())))
+            from datalad.distribution.dataset import resolve_path
+            container_path = resolve_path(container_name, ds)
+            container = [c for c in containers.values()
+                         if c['path'] == container_path]
+            if container:
+                container = container[0]
```
yarikoptic: If multiple match (shouldn't happen I know), better be explicit (don't run or issue a warning about detected ambiguity).

bpoldrack: Done

PR #54, opened 2018-10-24
BF: yield results in time

bpoldrack:
`containers-run` used to delay results from `run`. Fixed this.


bpoldrack:
Hehe  - reused from previous PR ;-)

PR #55, opened 2018-10-26
ENH: add: Copy local files instead of transforming them to a URL

### LINKED ISSUE #47: add: local path for --url fails with git-annex 6.20180626
As expected, following git-annex 6.20180626, running containers-add with a local url and no additional configuration will fail:

```bash
%> datalad containers-add foo --url /tmp/foo.simg
Configuration does not allow accessing file:///tmp/foo.simg
Configuration does not allow accessing file:///tmp/foo.simg
git-annex: addurl: 1 failed
[ERROR  ] CommandError: command 'addurl'
| Error, annex reported failure for addurl (url='file:///tmp/foo.simg'): {'command': 'addurl', 'success': False, 'file': None} [containers_add(/tmp/dl/another/.datalad/environments/foo/image)]
containers_add(error): /tmp/dl/another/.datalad/environments/foo/image (file) [CommandError: command 'addurl'
Error, annex reported failure for addurl (url='file:///tmp/foo.simg'): {'command': 'addurl', 'success': False, 'file': None}]
containers_add(error): /tmp/dl/another/.datalad/environments/foo/image (file) [no image at /tmp/dl/another/.datalad/environments/foo/image]
action summary:
  containers_add (error: 2)
```

git-annex's output to stderr mentions that it's a configuration issue [*], but from that information alone I doubt many DataLad users would know what to do next.  So I think we should do something here.  The options I can think of:

  * Give a more specific error message (at the level of containers_add or annexrepo.add_url_to_file)

    We could tell the user the do something like `git config annex.security.allowed-url-schemes file`, but we shouldn't be recommending that setting to users.  It'd be nicer if we could recommend a one-time override with `datalad -c annex.security.allowed-url-schemes=file ...`, but that setting isn't propagated to the git-annex call (and would fail on subsequent `git annex get` calls).

  * Stop supporting local files as an argument to --url

  * Add locals files via direct copying rather than an `add_url_to_file` call

    I think the main downside would be that the local source is no longer recorded by git-annex and we can't run `git annex get` on the file.

[*] I'm not sure why it does so twice; I guess add_url_to_file makes two addurl calls at some point, but I haven't checked.

kyleam:
As of git-annex 6.20180626, using a local file as the argument to
--url will fail with the default configuration for
annex.security.allowed-url-schemes.  Adding a local image is primarily
done for testing purposes, and pointing to a local URL doesn't provide
much of a benefit, so just copy it instead.  If a user wants to treat
a local path as a URL, they can still specify it with 'file://'.

Closes #47.

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -4,7 +4,9 @@ __docformat__ = 'restructuredtext'
 
 import re
 import logging
+import os
 import os.path as op
+from shutil import copyfile
 from simplejson import loads
 
 from datalad.interface.base import Interface
@@ -17,7 +19,6 @@ from datalad.support.constraints import EnsureStr
 from datalad.support.constraints import EnsureNone
 from datalad.support.exceptions import InsufficientArgumentsError
 from datalad.interface.results import get_status_dict
-from datalad.support.network import get_local_file_url
 
 # required bound commands
 from datalad.coreapi import save
@@ -30,12 +31,7 @@ lgr = logging.getLogger("datalad.containers.containers_add")
 def _resolve_img_url(url):
     """Takes a URL and tries to resolve it to an actual download
     URL that `annex addurl` can handle"""
-    if op.exists(url):
-        lgr.debug(
-            'Convert local path specification into a file:// URL')
-        # annex wants a real url
-        url = get_local_file_url(url)
-    elif url.startswith('shub://'):
+    if url.startswith('shub://'):
         lgr.debug('Query singularity-hub for image download URL')
         import requests
         req = requests.get(
@@ -187,6 +183,12 @@ class ContainersAdd(Interface):
                     docker_image, image)
                 check_call(["docker", "pull", docker_image])
                 docker.save(docker_image, image)
+            elif op.exists(url):
+                lgr.info("Copying local file %s to %s", url, image)
+                image_dir = op.dirname(image)
+                if image_dir and not op.exists(image_dir):
+                    os.makedirs(image_dir)
+                copyfile(url, image)
             else:
                 try:
                     # ATM gives no progress indication
--- a/datalad_container/tests/test_containers.py
+++ b/datalad_container/tests/test_containers.py
@@ -13,6 +13,8 @@ from datalad.tests.utils import ok_
 from datalad.tests.utils import assert_status
 from datalad.tests.utils import assert_raises
 from datalad.tests.utils import assert_result_count
+from datalad.tests.utils import assert_in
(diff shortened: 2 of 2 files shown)


datalad_container/containers_add.py
```diff
@@ -187,6 +183,12 @@ def __call__(name, url=None, dataset=None, call_fmt=None, image=None):
                     docker_image, image)
                 check_call(["docker", "pull", docker_image])
                 docker.save(docker_image, image)
+            elif op.exists(url):
+                lgr.info("Copying local file %s to %s", url, image)
+                image_dir = op.dirname(image)
+                if image_dir and not op.exists(image_dir):
+                    os.makedirs(image_dir)
+                copyfile(url, image)
```
yarikoptic: - possibly technically irrelevant since annex doesn't store even executable permission afaik for annexed files, but might manifest somehow many in in "unlocked v6 mode" -- `copyfile` doesn't copy permissions (image could be executable as e.g. in the case with singularity). `.copy2` seems to do `copyfile` and then `copystat`
- (could be a separate PR if too involved) is there a way to facilitate efficient copy on CoW filesystems (BTRFS)?    `cp --reflink=auto` on GNU systems, `cp -c` on OSX  achieves that in cmdline. For python quick google lead to a alpha/stale effort https://github.com/giampaolo/zerocopy which I didn't check inside (may be just does those cmdline calls?)

kyleam: > * possibly technically irrelevant since annex doesn't store even executable permission afaik for annexed files, but might manifest somehow many in in "unlocked v6 mode" -- `copyfile` doesn't copy permissions (image could be executable as e.g. in the case with singularity). `.copy2` seems to do `copyfile` and then `copystat`

Hmm, my thinking was that *not* preserving the executable bit was (1) consistent with the addurl behavior on the local file and (2) makes people configure `cmdexec` rather than rely on the executable bit, which may not be preserved across machines, at least until

https://github.com/datalad/datalad-container/blob/1983d6fbb3f76cf73e31b17eca5a27207e00eb2e/datalad_container/containers_add.py#L198-L199

I'm not strongly tied to the reasoning though (particularly since I view the `--url=<local-file>` behavior as mainly useful for local testing anyway), so I'm OK updating it to `copy2` if you'd prefer.

> * (could be a separate PR if too involved) is there a way to facilitate efficient copy on CoW filesystems (BTRFS)? [...]

Yeah, that seems outside the scope of this PR.

yarikoptic: Makes sense

PR #56, opened 2018-10-30
BF: Do not fail on not installed subdatasets

bpoldrack:
`container-list` used to fail on subdatasets that aren't present (empty dir). This fixes it.

### DIFF
--- a/datalad_container/containers_list.py
+++ b/datalad_container/containers_list.py
@@ -44,9 +44,11 @@ class ContainersList(Interface):
                              purpose='list containers')
 
         for sub in ds.subdatasets(return_type='generator'):
-            for c in Dataset(sub['path']).containers_list():
-                c['name'] = sub['gitmodule_name'] + '/' + c['name']
-                yield c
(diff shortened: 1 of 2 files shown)


datalad_container/containers_list.py
```diff
@@ -44,9 +44,11 @@ def __call__(dataset=None):
                              purpose='list containers')
 
         for sub in ds.subdatasets(return_type='generator'):
-            for c in Dataset(sub['path']).containers_list():
-                c['name'] = sub['gitmodule_name'] + '/' + c['name']
-                yield c
+            subds = Dataset(sub['path'])
+            if subds.is_installed():
```
yarikoptic: I wonder if it is worth collecting the list of those sub-datasets which aren't installed and issuing `lgr.info` at the end with the message like `"%d subdatasets which might contain containers we found not installed."`

bpoldrack: True. Will do. Once at it, I should also check `run-procedure --discover` for this issue and treat it similarly.

PR #60, opened 2018-11-30
 ENH: run: Move the image from inputs to extra_inputs 

### LINKED ISSUE #38: Adding image to inputs may surprise users that use "{inputs}" placeholder
Taking containers-run as a drop-in replacement for run, a user wouldn't expect to see the image at the end of the input list.

```bash
%> datalad containers-run "echo foo >foo"
%> datalad containers-run --input=foo "echo {inputs} | sed 's/ /\n/g' >input-list"
%> cat input-list
foo
.datalad/environments/bbox/image
```

We could add yet another argument (something like `inputs_extra`) to `run.run_command` and have containers-run use that directly.  I don't think that's pretty, but I can't think of anything else at the moment.

kyleam:
`containers-run` builds on `run`, augmenting the input list with a
Singularity image.  This means that if a user then puts "{inputs}" in
their command, the image file is included too.  This is almost
certainly surprising to the user because the user has not explicitly
specified the image with --input and is probably not aware that
containers-run adds it.  Avoid this by using run_command's recently
added extra_inputs argument.

This depends on datalad/datalad#3038.

Closes #38.

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -123,9 +123,6 @@ class ContainersRun(Interface):
             # just prepend and pray
             cmd = container['path'] + ' ' + cmd
 
-        # with amend inputs to also include the container image
-        inputs = (inputs or []) + [image_path]
-
         with patch.dict('os.environ',
                         {CONTAINER_NAME_ENVVAR: container['name']}):
             # fire!
@@ -133,6 +130,7 @@ class ContainersRun(Interface):
                     cmd=cmd,
                     dataset=dataset or (ds if ds.path == pwd else None),
                     inputs=inputs,
+                    extra_inputs=[image_path],
                     outputs=outputs,
                     message=message,
                     expand=expand,
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -15,6 +15,7 @@ from datalad.interface.utils import eval_results
 
 from datalad.interface.results import get_status_dict
 from datalad.interface.run import Run
+from datalad.interface.run import run_command
 from datalad.interface.run import get_command_pwds
 from datalad.interface.run import normalize_command
 from datalad_container.find_container import find_container
@@ -128,7 +129,7 @@ class ContainersRun(Interface):
         with patch.dict('os.environ',
                         {CONTAINER_NAME_ENVVAR: container['name']}):
             # fire!
-            for r in Run.__call__(
+            for r in run_command(
                     cmd=cmd,
                     dataset=dataset or (ds if ds.path == pwd else None),
                     inputs=inputs,
@@ -136,7 +137,5 @@ class ContainersRun(Interface):
                     message=message,
                     expand=expand,
                     explicit=explicit,
-                    sidecar=sidecar,
-                    on_failure="ignore",
-                    return_type='generator'):
+                    sidecar=sidecar):
                 yield r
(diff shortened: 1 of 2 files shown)


yarikoptic:
We shouldn't forget to boost version dependency on datalad before releasing this one

yarikoptic:
FWIW #3038 is merged.  I thought we merged this one as well and thought to use it.

yarikoptic:
I think I ran into this issue again, forgot that I probably installed from your PR branch, then upgraded to current master, and kaboom ... took some time to arrive back to this PR to recall details - downgraded to your PR branch and now it works again.
we should definitely finish up/merge this PR  asap, restarted travis builds

kyleam:
> we should definitely finish up/merge this PR asap, restarted travis builds

I'll leave merging up to you.

I'm hesitant to merge it because that means it might be released, along with the required datalad version bump, before there is a fix for datalad/datalad#3087 (fallout from datalad/datalad#3009), which breaks the "dhub://" prefix.  It looks like the current plan for 3087 is to wait for revolution pr to be merged (datalad/datalad#3106), though IMO 3087 is a bigger issue than original issues addressed by 3009, and it'd be better to revert the changes from 3009 while we wait for the revolution pr.

yarikoptic:
although i failed to fix that https://github.com/datalad/datalad/issues/3087 yet (may be revolution would just fix it all!) is this change anyhow directly demands it?  if not, I would say we should merge -- I keep needing this fix, and haven't used `dhub://` yet -- all images I use are singularity.

kyleam:
>  is this change anyhow directly demands it?

I don't have anything to add to what I said above.

kyleam:
Rebased and bumped DataLad dependency in setup.py.

<details>
<summary>range-diff</summary>

```
1:  70fe020 ! 1:  d2dd8d7 RF: run: Use run_command()
    @@ -17,7 +17,7 @@
     +from datalad.interface.run import run_command
      from datalad.interface.run import get_command_pwds
      from datalad.interface.run import normalize_command
    - from datalad_container.containers_list import ContainersList
    + from datalad_container.find_container import find_container
     @@
              inputs = (inputs or []) + [image_path]
      
    @@ -25,7 +25,7 @@
     -        for r in Run.__call__(
     +        for r in run_command(
                      cmd=cmd,
    -                 dataset=ds,
    +                 dataset=dataset or (ds if ds.path == pwd else None),
                      inputs=inputs,
     @@
                      message=message,
2:  7e03c5e ! 2:  89e9f03 ENH: run: Move the image from inputs to extra_inputs
    @@ -26,7 +26,7 @@
              # fire!
              for r in run_command(
                      cmd=cmd,
    -                 dataset=ds,
    +                 dataset=dataset or (ds if ds.path == pwd else None),
                      inputs=inputs,
     +                extra_inputs=[image_path],
                      outputs=outputs,
-:  ------- > 3:  2036812 MNT: setup.py: Bump require DataLad version for extra_inputs
```

</details>

yarikoptic:
https://github.com/datalad/datalad/pull/3106 is ready for review to address the issue with adding directories.  Unless some show stopped is found, it should make it into datalad 0.11.5

yarikoptic:
I kicked 0.11.5 out -- so after conflicts resolved we could merge/release!

kyleam:
Didn't realize there was yet another round of conflicts.  Will update.

PR #62, opened 2018-12-13
 ENH: add: Support Singularity's "docker://" URI

kyleam:
Ideally a user is able to use the shub:// scheme so that we resolve a
URL to register with git-annex, but, for images that are only
available from Docker, it is still useful to support Singularity's
"docker://" scheme.

Sadly, this introduces the possibility for confusion between the
"docker://" and "dhub://" prefix, but it seems worth it in order to
support pulling in Docker images via Singularity.

---

- [x] tests for adding container with "docker://" prefix

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -9,6 +9,7 @@ import os.path as op
 from shutil import copyfile
 from simplejson import loads
 
+from datalad.cmd import Runner
 from datalad.interface.base import Interface
 from datalad.interface.base import build_doc
 from datalad.support.param import Parameter
@@ -136,6 +137,7 @@ class ContainersAdd(Interface):
 
         ds = require_dataset(dataset, check_installed=True,
                              purpose='add container')
+        runner = Runner()
 
         # prevent madness in the config file
         if not re.match(r'^[0-9a-zA-Z-]+$', name):
@@ -216,14 +218,13 @@ class ContainersAdd(Interface):
             lgr.debug('Attempt to obtain container image from: %s', imgurl)
             if url.startswith("dhub://"):
                 from .adapters import docker
-                from subprocess import check_call
 
                 docker_image = url[len("dhub://"):]
 
                 lgr.debug(
                     "Running 'docker pull %s and saving image to %s",
                     docker_image, image)
-                check_call(["docker", "pull", docker_image])
+                runner.run(["docker", "pull", docker_image])
                 docker.save(docker_image, image)
             elif op.exists(url):
                 lgr.info("Copying local file %s to %s", url, image)
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -52,7 +52,7 @@ def _guess_call_fmt(ds, name, url):
     """
     if url is None:
         return None
-    elif url.startswith('shub://'):
+    elif url.startswith('shub://') or url.startswith('docker://'):
         return 'singularity exec {img} {cmd}'
     elif url.startswith('dhub://'):
         return 'python -m datalad_container.adapters.docker run {img} {cmd}'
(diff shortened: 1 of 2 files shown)


yarikoptic:
Thank you @kyleam ! I have managed to miss this PR.

Could we add a basic unittest for that call with e.g. docker://busybox? (docker image is just 1.2M and singularity image I got was just 800k)

kyleam:
> Could we add a basic unittest for that call with e.g. docker://busybox?

Yep, I plan to revisit this and add a test (as the checkbox in the initial post indicates :] ).

yarikoptic:
> > Could we add a basic unittest for that call with e.g. docker://busybox?
> 
> Yep, I plan to revisit this and add a test (as the checkbox in the initial post indicates :] ).

sorry -- missed the checkbox I think because of the divider line which gave impression of "the description is over" and I got used to have those more inline within the description.  Now will keep an eye to such arrangements as well

kyleam:
> I think because of the divider line which gave impression of "the description is over" 

FWIW, for one-commit PRs, I've been using the divider line to show where the commit message ends and extra PR text begins.

yarikoptic:
> Sadly, this introduces the possibility for confusion between the
> "docker://" and "dhub://" prefix, but it seems worth it in order to
> support pulling in Docker images via Singularity.

I wondered if we should make it more explicit right from the beginning... who knows what other container technologies will come and be able to import from docker:// etc.  So may be worthwhile adding explicit `--type` (or `--backend`)  with default `"auto"` (or None, to signal "automatic") so we could use `add docker://` to add a docker image for docker, and `add --type singularity docker://` to give it to singularity backend to handle?
More magical way would be to adopt git (or pip) strategy to prefix with "backend" separated from original url with `::` (or `+` for pip), so would become `add singularity::docker://`. (ref: in git it is e.g. `git clone hg+https://...`; and for pip `pip install git+git://...` ;-))

kyleam:
Updated with test:


```
% git range-diff master kyleam/enh-docker-to-sing enh-docker-to-sing     
1:  9706d3f = 1:  2587845 RF: add: Replace check_call() with Runner   
2:  9314397 ! 2:  4b8e939 ENH: add: Support Singularity's "docker://" URI
    @@ -58,3 +58,42 @@
                  elif op.exists(url):
                      lgr.info("Copying local file %s to %s", url, image)
                      image_dir = op.dirname(image)
    +
    + diff --git a/datalad_container/tests/test_schemes.py b/datalad_container/tests/test_schemes.py
    + new file mode 100644
    + --- /dev/null
    + +++ b/datalad_container/tests/test_schemes.py
    +@@
    ++import os.path as op
    ++
    ++from datalad.api import Dataset
    ++from datalad.api import create
    ++from datalad.api import containers_add
    ++from datalad.api import containers_list
    ++from datalad.api import containers_run
    ++
    ++from datalad.tests.utils import ok_clean_git
    ++from datalad.tests.utils import ok_file_has_content
    ++from datalad.tests.utils import assert_result_count
    ++from datalad.tests.utils import with_tempfile
    ++from datalad.tests.utils import skip_if_no_network
    ++
    ++
    ++@skip_if_no_network
    ++@with_tempfile
    ++def test_docker(path):  # Singularity's "docker://" scheme.
    ++    ds = Dataset(path).create()
    ++    ds.containers_add(
    ++        "bb",
    ++        url=("docker://busybox@sha256:"
    ++             "7964ad52e396a6e045c39b5a44438424ac52e12e4d5a25d94895f2058cb863a0"))
    ++
    ++    img = op.join(ds.path, ".datalad", "environments", "bb", "image")
    ++    assert_result_count(ds.containers_list(), 1, path=img, name="bb")
    ++    ok_clean_git(path)
    ++
    ++    ds.containers_run(["sh", "-c", "cat /singularity >out"])
    ++    ok_file_has_content(
    ++        op.join(ds.path, "out"),
    ++        "There is no runscript defined for this container",
    ++        re_=True, match=False)

```

kyleam:
> I wondered if we should make it more explicit right from the beginning... 

Hmm, I view both your `--type` and extended scheme suggestions as an overkill at the moment, but if others want to be more forward-thinking than I am, I'll go along :]

PR #65, opened 2019-01-31
run: Check name for slash before searching subdatasets

### LINKED ISSUE #64: containers-* commands  takes awhile since interrogates full list of subdatasets
Since https://github.com/datalad/datalad-container/pull/51 we by default list all containers even the ones discovered in subdatasets.  This takes awhile in sizeable datasets hierarchy.  I think we should just add `-r|--recursive` flag and interrogate subdatasets only if it is specified.  Would make behavior more consistent

kyleam:
This follows up on discussion from #64 and implements the "recurse if / is in the name" proposal.

@bpoldrack mentions there that

> If we are going to do that, we could also consider to just "parse" the name and go down that path instead of triggering a "full" recursion.

I think that's a good idea and something we should probably do at some point, but the PR punts on that and goes with the simpler "all or nothing" approach.

The main commits of interest are

  * ENH: list: Don't recurse subdatasets by default
  * ENH: run: Look for containers in subdatasets only if name has "/"

Closes #64.

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -15,7 +15,7 @@ from datalad.interface.utils import eval_results
 from datalad.interface.run import Run
 from datalad.interface.run import get_command_pwds
 from datalad.interface.run import normalize_command
-from datalad_container.containers_list import ContainersList
+from datalad_container.find_container import find_container
 
 lgr = logging.getLogger("datalad.containers.containers_run")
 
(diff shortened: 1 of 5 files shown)


datalad_container/containers_run.py
```diff
@@ -37,9 +37,11 @@ class ContainersRun(Interface):
     # the rest is put in the verbose help and manpage
     """Drop-in replacement of 'run' to perform containerized command execution
 
-    Container(s) need to be configured beforehand (see containers-add).
-    If only one container is known, it will be selected automatically,
-    otherwise a specific container has to be specified.
+    Container(s) need to be configured beforehand (see containers-add). If no
+    container is specified and only one container is configured in the
+    top-level dataset, it will be selected automatically. If more than one
```
yarikoptic: "top-level" (which I read as a top most superdataset in current hierarchy) or "current"?

datalad_container/find_container.py
```diff
@@ -0,0 +1,80 @@
...
+
+    for fn in fns:
+        container = fn(ds, container_name, containers)
+        if container:
+            return container
+
+    raise ValueError(
+        'Container selection impossible: not specified, ambiguous '
+        'or unknown (known containers are: {})'
+        ''.format(list(containers.keys()))
```
yarikoptic:
```suggestion
        ''.format(list(', '.join(containers)))
```

datalad_container/find_container.py
```diff
@@ -0,0 +1,80 @@
...
+                                                   recursive=recurse)}
+
+    if not containers:
+        raise ValueError("No known containers. Use containers-add")
+
+    fns = [
+        _get_the_one_and_only,
+        _get_container_by_name,
+        _get_container_by_path,
+    ]
```
yarikoptic: so it doesn't even bother with the one and only if not needed.  Although not yet directly relevant here, if the `containers` "thing" remains an iterator (not a dictionary, eventually), then we will be able to allow for early return even when exploring recursively (which could be lengthy). ATM it first does full traversal 

```suggestion
    fns = [] if container_name else [_get_the_one_and_only]
    fns += [
        _get_container_by_name,
        _get_container_by_path,
    ]
```

datalad_container/containers_run.py
kyleam: > "top-level" (which I read as a top most superdataset in current hierarchy)

OK, I'll changed it to "current".

datalad_container/find_container.py
kyleam: This is a pure copy from what was there before this PR, but I'll update it to your suggestion.

datalad_container/find_container.py
kyleam: > so it doesn't even bother with the one and only if not needed

Sorry, I don't see the suggested change as an improvement.

> Although not yet directly relevant here, if the `containers` "thing" remains an iterator (not a dictionary, eventually) [...]

The implementation here is staying close to the previous approach.  Reworking it further is beyond the scope of what I was intending to do here.  (Actually, I could even take or leave the last commit.  It can be safely dropped.)  If I did spend more time on reworking the approach, I'd prefer to focus on the more specific lookup approach suggested by Ben and mentioned in a223bad's message.

PR #68, opened 2019-03-05
BF: containers-list recursion

bpoldrack:
Found an (actually quite obvious) bug in containers-list. Recursion isn't really a recursion ATM and discovers containers only one level deep. Will push the fix as soon as the enhanced test showed how it fails.


bpoldrack:
Well, I somehow managed to not review #65, which I definitely should have. So, I consider this a failure on my side ;-)

PR #74, opened 2019-03-21
RF: do not depend on rc of datalad and scrapy which causes PIP to always consider RCs of them

yarikoptic:
See https://github.com/pypa/pip/issues/4969 for more info/"rationale"

Similar change was done to -crawler already


yarikoptic:
I think it was a nobrainer

Issue #75, opened 2019-04-19
More meaningful error message(s) whenever cmdexec contains "unknown" placeholders


yarikoptic:
```shell
$> datalad containers-list
containers(ok): /home/yoh/proj/repronim/containers/singularity/bids/bids-validator--1.2.3 (file)

$> datalad containers-run 
[ERROR  ] u'dspath' [containers_run.py:__call__:87] (KeyError) 
```
since I thought (from `datalad run --help`) that `dspath` is available.  But I think it would be better to provide custom error message like

     `cmdexec` references an unknown placeholder `dspath`

or alike.
Also, it might be nice to verify (if possible) all those placeholders in `containers-add` whenever a custom `--call-fmt` was provided?

kyleam:
@yarikoptic:

> $> datalad containers-run 
> [ERROR  ] u'dspath' [containers_run.py:__call__:87] (KeyError) 
> ```
> since I thought (from `datalad run --help`) that `dspath` is
> available.

The placeholders accepted by datalad-run are unrelated to the cmdexec
value stored in .datalad/config.  The placeholders in --call-fmt/cmdexec
are formatted by containers-run to generate the command it passes to
datalad-run.  As described by containers-add's help, the two
placeholders currently supported are "{img}" and "{cmd}".  If you want a
placeholder in cmdexec to make it through to datalad-run, you'd have to
put another layer of curly brackets around it.

> But I think it would be better to provide custom error message like
>
>      `cmdexec` references an unknown placeholder `dspath`
>
> or alike.

I agree that it'd be better for containers-run to catch the key error
and provide a more meaningful message (or ideally a result).
datalad-run already does so for its placeholders:

  $ datalad run "echo {dsspath}"
  run(impossible): /tmp/dlc-yk2LIS2 (dataset) [command has an unrecognized placeholder: 'dsspath']

I'll adjust containers-run to do similar.

> Also, it might be nice to verify (if possible) all those placeholders in `containers-add` whenever a custom `--call-fmt` was provided?

It is possible.  I wouldn't be opposed to a patch to do so, though I
personally think doing the check in containers-run is good enough and
maybe preferable so that the logic is in just one spot.

Issue #72, opened 2019-03-14
incorrect operation of containers-run from within subdirectories


yarikoptic:
```shell
smaug:~/proj/nuisance$ cat .datalad/config
[datalad "dataset"]
        id = 1a495a66-b20d-11e8-9335-002590f97d84
[datalad "containers.reproin"]
        updateurl = shub://ReproNim/reproin:0.5.3
        image = .datalad/environments/reproin/image
        cmdexec = singularity exec {img} {cmd}
[datalad "containers.mriqc"]
        image = .datalad/environments/mriqc/image
        cmdexec = .datalad/environments/_bin/singularity-exec {img} {cmd}
[datalad "containers.simple-workflow"]
        image = .datalad/environments/simple-workflow/image
        cmdexec = singularity exec {img} {cmd}
```
and the shim is
```shell
smaug:~/proj/nuisance$ cat .datalad/environments/_bin/singularity-exec
#!/bin/bash

# A shim to take care about running sanitized home etc

bindir=$(dirname $0 | xargs readlink -f)
#echo "I: bindir=$bindir"
homedir=$(echo $bindir | xargs dirname )/_home
topdir=$(echo $bindir | xargs dirname | xargs dirname | xargs dirname )

echo "I: exec'ing with sanitized singularity with"
echo "I:   HOME=$homedir"
echo "I:   $topdir bind mounted"
singularity exec \
        -H "$homedir" \
        -B "$topdir:$topdir" \
        -e \
        "$@"
```

if I cd under `code/` subdirectory `containers-run` fails in a number of ways:
```shell
smaug:~/proj/nuisance/code$ datalad containers-run  -n mriqc --explicit pwd
[INFO   ] Making sure inputs are available (this may take some time) 
[WARNING] path not associated with any dataset [get(/home/chrispycheng/proj/.datalad/environments/mriqc/image)] 
[WARNING] Input does not exist: /home/chrispycheng/proj/.datalad/environments/mriqc/image 
[INFO   ] == Command start (output follows) ===== 
I: exec'ing with sanitized singularity with
I:   HOME=/home/chrispycheng/proj/nuisance/.datalad/environments/_home
I:   /home/chrispycheng/proj/nuisance bind mounted
ERROR  : Image path ../.datalad/environments/mriqc/image doesn't exist: No such file or directory
ABORT  : Retval = 255
[INFO   ] == Command exit (modification check follows) ===== 
[...]

PR #76, opened 2019-04-19
ENH: {dspath} to point to the dataset containing the container

yarikoptic:
Relates to https://github.com/datalad/datalad-container/issues/75 where I
blissfully forgot about the "sandwich" of placeholders where some are
evalutated by datalad_containers (ATM only {img} and {cmd}) and the ones
to be passed into datalad-run should be double-wrapped with {}.

Since {dspath} of the dataset where actual execution to be done is of no
concern for variables at containers-run level, I have decided to expose the
(relative) path to it within {dspath}.  containers-list would return "dspath"
record with the full path to it, and then (similarly to image) it would get
converted to the relative path from current ({pwd} for datalad-run) directory.
I have not tried to "rerun" such runs, but it seems to work at least at the
"containers-run" level.

Motivation:

In my case I want to use a custom "runner" script, which is kept in the same
repository as the containers. So I would like to have a variable I could use to
refer to the top of the dataset which contains containers (which might be
different from the dataset where I actually execute, e.g. when dataset with
containers is a subdataset of the outside container).  It also relates but
orthogonal to the #72 (`containers-run` within a subdirectory), where the
[proposed patch](https://github.com/datalad/datalad-container/issues/72#issuecomment-472925070)
seems to address the issue for running within subdirectory and on my trial
plays nicely with the change proposed here -- I was able to
containers-run a container from within a subdataset within a
sub-directory of the super-dataset.

This PR ATM comes without a unit-test since first I thought to seek feedback
from datalad-containers/run gurus.

### DIFF
--- a/datalad_container/containers_list.py
+++ b/datalad_container/containers_list.py
@@ -75,6 +75,8 @@ class ContainersList(Interface):
             if 'image' not in v:
                 # there is no container location configured
                 continue
+            if dataset and 'dspath' not in v:
+                v['dspath'] = dataset.path
             res = get_status_dict(
                 status='ok',
                 action='containers',
--- a/datalad_container/containers_list.py
+++ b/datalad_container/containers_list.py
@@ -75,14 +75,13 @@ class ContainersList(Interface):
             if 'image' not in v:
                 # there is no container location configured
                 continue
-            if dataset and 'dspath' not in v:
-                v['dspath'] = dataset.path
             res = get_status_dict(
                 status='ok',
                 action='containers',
                 name=k,
                 type='file',
                 path=op.join(ds.path, v.pop('image')),
+                parentds=ds.path,
                 # TODO
                 #state='absent' if ... else 'present'
                 **v)
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -105,7 +105,9 @@ class ContainersAdd(Interface):
             doc="""Command format string indicating how to execute a command in
             this container, e.g. "singularity exec {img} {cmd}". Where '{img}'
             is a placeholder for the path to the container image and '{cmd}' is
-            replaced with the desired command.""",
+            replaced with the desired command. Additional placeholders:
+            '{img_dspath}' is relative path to the dataset containing the image.
+            """,
             metavar="FORMAT",
             constraints=EnsureStr() | EnsureNone(),
         ),
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -62,6 +62,9 @@ class ContainersRun(Interface):
 
         container = find_container(ds, container_name)
         image_path = op.relpath(container["path"], pwd)
+        # container record would contain path to the (sub)dataset containing
+        # it.  If not - take current dataset, as it must be coming from it
+        container_dspath = op.relpath(container.get('dspath', ds.path), pwd)
 
         # sure we could check whether the container image is present,
         # but it might live in a subdataset that isn't even installed yet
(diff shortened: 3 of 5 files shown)


kyleam:
@yarikoptic:

> Since {dspath} of the dataset where actual execution to be done is of no
> concern for variables at containers-run level, I have decided to expose the
> (relative) path to it within {dspath}.  containers-list would return "dspath"
> record with the full path to it, and then (similarly to image) it would get
> converted to the relative path from current ({pwd} for datalad-run) directory.

I think I understand your use case, and I can't think of a better way to
do it.  I think we shouldn't call it dspath, though, to avoid possible
confusion with datalad-run's dspath.  Maybe "img_dspath"?

yarikoptic:
yeah, I was/still debating either it should have been `{dspath}` or something else. I think making it more explicit (`{img_dspath}` sounds good to me, aligns with `{img}`) would indeed be better.  Let me then push it "forward" with that in mind.

yarikoptic:
I am not quite sure why my mocking in 9b32aba00f80c581c3c35016d71c9180bba0b336 seems to be of no effect while tests pass locally... so I will just revert it (while keeping in history as a reminder that like could/should be faster ;))

datalad_container/containers_list.py
```diff
@@ -75,6 +75,8 @@ def __call__(dataset=None, recursive=False):
             if 'image' not in v:
                 # there is no container location configured
                 continue
+            if dataset and 'dspath' not in v:
+                v['dspath'] = dataset.path
```
kyleam: I think this would be better as

```diff
diff --git a/datalad_container/containers_list.py b/datalad_container/containers_list.py
index cda137c..b08eff0 100644
--- a/datalad_container/containers_list.py
+++ b/datalad_container/containers_list.py
@@ -75,13 +75,12 @@ def __call__(dataset=None, recursive=False):
             if 'image' not in v:
                 # there is no container location configured
                 continue
-            if dataset and 'dspath' not in v:
-                v['dspath'] = dataset.path
             res = get_status_dict(
                 status='ok',
                 action='containers',
                 name=k,
                 type='file',
+                parentds=ds.path,
                 path=op.join(ds.path, v.pop('image')),
                 # TODO
                 #state='absent' if ... else 'present'
```

Changes:

  * Drop condition on `dataset` and instead use resolved `ds`.  Otherwise the value won't be set for images in the top-level dataset if an explicit dataset isn't supplied.
  * Don't bother checking for the value in config dict because (1) we don't save this value and (2) I can't see why the user should be able to manually override this value in the config.
  * Call the field `parentds` for consistency with the field in `status`.

Thoughts?

yarikoptic: Thanks for the review/suggestions!  Re `parentds` - that seems would make it also consistent with `metadata`'s records.  
I wondered if then `img_dspath` should be longer/uglier `img_parentdspath` for having clear association. But since we are abbreviating anyways, I guess we will just consider `dspath` to be ok there.
Anyways -- will need to adjust `containers-add --help` to mention this new placeholder.

yarikoptic:
> They say "do not use any real container", but a real container isn't used in either case.

Thanks, indeed.  Reworded to be "do not require any real container" -- since it is still required AFAIK although indeed not used.

> combining the test with the code change

I have got into the habit of introducing two separate commits so I could quickly revert the code change and see if added after test fails.  Also it makes easy to do the opposite - to pick up a test to another branch and see if fails.  Not that I do that often, but I think that was the reason why I often do it that way.

kyleam:
Thanks for the update.  Looks good to me.  (Pushed a minor fixup to a test.)

@yarikoptic:

>> They say "do not use any real container", but a real container isn't used in either case.
>
> Thanks, indeed.  Reworded to be "do not require any real container" -- since it is still required AFAIK although indeed not used.

I might be getting myself confused, but AFAICS the code in question
didn't involve a real container in the first place.  396d42d (RF(OPT):
do not require any real container, 2019-04-19) changes code that
involved a file some_container.img with the contents "doesn't matter".
At any rate, it's fine as is; I've understood from the beginning that
polishing topic branches isn't common practice in datalad projects.

>> combining the test with the code change
>
> I have got into the habit of introducing two separate commits so I could quickly revert the code change and see if added after test fails.  Also it makes easy to do the opposite - to pick up a test to another branch and see if fails.  Not that I do that often, but I think that was the reason why I often do it that way. 

Yeah, I've seen and understand that reasoning for separating code/test
changes.  I haven't found it useful for my personal workflow because I
find it easy enough to revert just the code changes from a single commit
for local testing, but it's not a big deal either way.

yarikoptic:
Thanks for the tests fixup - I will resolve conflicts now and push again

PR #79, opened 2019-04-23
MNT: Drop Docker adapter

kyleam:
It was a useful idea to explore, but it's unlikely that anyone uses
it.  When an image is only available from Docker, we now support the
"docker://" prefix for building Singularity images, so it's not clear
that this is worth maintaining.  (And, although not the adapter's
fault, its existence sadly holds up the merge of gh-60.)  If there is
something that "docker://" won't work for, we can always resurrect
this.


yarikoptic:
Oh, let's keep the beast! It is not it's fault that it might be underused ;-) I will try to fix that issue ASAP over way it another

kyleam:
It's not just that it isn't used: it is probably redundant.  Again, nothing stops us from **easily** resurrecting it.

yarikoptic:
Well, the "problem" with singularity is that it is Linux only, unlike docker one way or another supported on OSX and Windows.  In the realm of https://github.com/ReproNim/containers I even thought, but never "formalized" until [now](https://github.com/ReproNim/containers/issues/1) the idea of having also docker images in the same repo.

kyleam:
> never "formalized" until [now](https://github.com/ReproNim/containers/issues/1) the idea of having also docker images in the same repo

I suspect you'll quickly find that this adapter was never more than an unpolished prototype.

PR #82, opened 2019-05-07
RF: small adaptions to current datalad

bpoldrack:


kyleam:
-containers should be compatible with the developmental DataLad, but I don't see why we should make it the minimum requirement.

bpoldrack:
Although I agree wrt functionality - with the ongoing integration of revolution just being compatible would mean to write different code for different versions of datalad. With an old version of datalad we would need to require revolution in addition and therefore have different imports. I'd rather just keep updating the version of datalad dependency until revolution is done.

PR #84, opened 2019-05-15
ENH: set DATALAD_CONTAINER_NAME env var while running a container

yarikoptic:
The main reason was due to the fact that older (prior 3.0?) singularity
followed the symlinked for the annexed file to set
SINGULARITY_NAME and SINGULARITY_CONTAINER env vars:

    $> singularity --version ; singularity exec images/repronim/repronim-reproin--0.5.4.sing bash -c 'export | grep SING'
    2.6.1-dist
    declare -x SINGULARITY_CONTAINER="MD5E-s301797407--a50bf88c0fb53809154e42e8d08a482e.4.sing"
    declare -x SINGULARITY_NAME="MD5E-s301797407--a50bf88c0fb53809154e42e8d08a482e.4.sing"

That seems was fixed in 3.x;

	$> singularity --version ; singularity exec images/repronim/repronim-reproin--0.5.4.sing bash -c 'export | grep SING'
	singularity version 3.0.3+ds
	declare -x SINGULARITY_APPNAME=""
	declare -x SINGULARITY_CONTAINER="/home/yoh/proj/repronim/containers/images/repronim/repronim-reproin--0.5.4.sing"
	declare -x SINGULARITY_NAME="repronim-reproin--0.5.4.sing"

but

- in many places 2.x is still be used
- our container name but default does not even match container
  name and just called "image"
- even when it is set to some reasonable named file, name of the
  container could still be different

So I thought that it should only be advantageous to expose our container name
to underlying process, possibly a shim, to take advantage of it, e.g. to set
PS1 env variable for interactive shell etc.

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -3,6 +3,7 @@
 __docformat__ = 'restructuredtext'
 
 import logging
+import os
 import os.path as op
 
 from datalad.interface.base import Interface
@@ -20,6 +21,10 @@ from datalad_container.find_container import find_container
 
 lgr = logging.getLogger("datalad.containers.containers_run")
 
+# Environment variable to be set during execution to possibly
+# inform underlying shim scripts about the original name of
+# the container
+CONTAINER_NAME_ENVVAR = 'DATALAD_CONTAINER_NAME'
 
 _run_params = dict(
     Run._params_,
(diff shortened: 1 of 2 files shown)


datalad_container/containers_run.py
```diff
@@ -112,6 +124,9 @@ def __call__(cmd, container_name=None, dataset=None,
         # with amend inputs to also include the container image
         inputs = (inputs or []) + [image_path]
 
+        old_envvar = os.environ.get(CONTAINER_NAME_ENVVAR, None)
+        os.environ[CONTAINER_NAME_ENVVAR] = container['name']
```
kyleam: What about using patch.dict instead?  Right now the value is cleaned up only if something was already set and wouldn't be restored properly for python callers if there is an exception raised with `Run.__call__()`.

yarikoptic: Thank you @kyleam for review/suggestions. I am not sure why I was lazy to not use `patch.dict`. Did it (with delayed import, takes too long to import, at least while I am also Zoom'ing). Committed via `--amend`, pushed.

PR #86, opened 2019-05-21
BF: report "Update " in commit message only if an existing image was actually updated

yarikoptic:
I tend to use --update while adding or updating existing image.  That leads to
IMHO erroneous commit message that an image was updated whenever it was freshly
added/configured.  Docstring for --update says "If a container with `name` does
not already exist, this option is ignored.", so reporting Conffigure in the
cases when it was "ignored" is IMHO a correct thing to do

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -208,8 +208,10 @@ class ContainersAdd(Interface):
         # collect bits for a final and single save() call
         to_save = []
         imgurl = url
+        was_updated = False
         if url:
             if update and op.lexists(image):
+                was_updated = True
                 # XXX: check=False is used to avoid dropping the image. It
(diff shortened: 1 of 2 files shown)


kyleam:
> yarikoptic requested a review from kyleam

FWIW my preference is to not be spammed with review requests unless it is somehow serving a purpose (e.g., it's something you'd think I wouldn't look at otherwise or it's been X amount of time and you think it slipped my notice).

yarikoptic:
> > yarikoptic requested a review from kyleam
> 
> FWIW my preference is to not be spammed with review requests unless it is somehow serving a purpose (e.g., it's something you'd think I wouldn't look at otherwise or it's been X amount of time and you think it slipped my notice).

yes Sir, will (not) do Sir!

Issue #85, opened 2019-05-21
2 tests fail locally: test_add_noop and test_docker


yarikoptic:
```shell
$ python -m nose -s -v datalad_container
...
======================================================================
ERROR: datalad_container.tests.test_containers.test_add_noop
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/yoh/proj/datalad/datalad-master/datalad/tests/utils.py", line 607, in newfunc
    return t(*(arg + (filename,)), **kw)
  File "/home/yoh/proj/datalad/datalad-container/datalad_container/tests/test_containers.py", line 42, in test_add_noop
    on_failure='ignore')
  File "/home/yoh/proj/datalad/datalad-master/datalad/distribution/dataset.py", line 527, in apply_func
    return f(**kwargs)
  File "/home/yoh/proj/datalad/datalad-master/datalad/interface/utils.py", line 490, in eval_func
    return return_func(generator_func)(*args, **kwargs)
  File "/home/yoh/proj/datalad/datalad-master/datalad/interface/utils.py", line 478, in return_func
    results = list(results)
  File "/home/yoh/proj/datalad/datalad-master/datalad/interface/utils.py", line 427, in generator_func
    result_renderer, result_xfm, _result_filter, **_kwargs):
  File "/home/yoh/proj/datalad/datalad-master/datalad/interface/utils.py", line 520, in _process_results
    for res in results:
  File "/home/yoh/proj/datalad/datalad-container/datalad_container/containers_add.py", line 249, in __call__
    copyfile(url, image)
  File "/usr/lib/python2.7/shutil.py", line 97, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 13] Permission denied: '/home/yoh/.tmp/datalad_temp_test_add_noopcjESpu/dummy'

======================================================================
FAIL: datalad_container.tests.test_schemes.test_docker
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/yoh/proj/datalad/datalad-master/datalad/tests/utils.py", line 841, in newfunc
    return func(*args, **kwargs)
  File "/home/yoh/proj/datalad/datalad-master/datalad/tests/utils.py", line 607, in newfunc
[...]

PR #87, opened 2019-05-21
BF: even more bogus URL + remove image before replacing with a new one

yarikoptic:
Partially addresses (the _noop test) #85 

The 2nd one is questionable, so up for discussion -- could be dropped.
more bogus url is not ultimate protection, but I didn't want to `chpwd`

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -246,6 +246,9 @@ class ContainersAdd(Interface):
                 image_dir = op.dirname(image)
                 if image_dir and not op.exists(image_dir):
                     os.makedirs(image_dir)
+                if op.lexists(image):
+                    lgr.debug("Image file %s already exists, removing", image)
+                    os.unlink(image)
                 copyfile(url, image)
             else:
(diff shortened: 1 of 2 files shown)


kyleam:
> The 2nd one is questionable, so up for discussion -- could be dropped.

I'm in favor of keeping the original behavior, but, as you say in #85, it should get a more informative message.

yarikoptic:
ok, I will then manually merge only the first commit into the master so we don't wait for the 2nd one

kyleam:
> ok, I will then manually merge only the first commit into the master so we don't wait for the 2nd one

Great, thanks!

yarikoptic:
I guess the second portion (removing image first) is to be abandoned, so this could be closed for now?

kyleam:
> I guess the second portion (removing image first) is to be abandoned, so this could be closed for now?

I don't consider it related to this PR, so, yes, my preference is for this to be closed, but I left it up to you since you are the one that first bundled the two and the one that did that manual merge.

PR #88, opened 2019-05-21
TST: Don't check content of /singularity

kyleam:
The content changes between Singularity v2 and v3.  This check was
only intended to be a loose check to see if it looks like we can
successfully run within the container, so instead check that 'ls
/singularity' doesn't fail.

Re: #85


yarikoptic:
Thanks!

Issue #91, opened 2019-07-08
Re-enable tests when singularity hub is back


yarikoptic:
ATM datalad-containers is all red on Travis. @vsoch [took the hub offline since there was some malicious user activity](https://twitter.com/vsoch/status/1146801275843072001).  I wonder if in such cases (`requests.exceptions.ConnectionError ... Failed to establish a new connection` ) we should just skip the test or keep them failing as our indicator of some disturbance in the force?

vsoch:
Singularity Hub being down shouldn't be a regular thing, but it will be down for now until further notice as we address the issue. For my repos that use shub endpoints, I either changed to Docker, or in the case that it was a specific image, just disabled for now. Not much more to do.

bpoldrack:
I think we should keep failing. Nobody will notice a skip, but sometimes we might want to do something about it and even if not so: The "passing" of the Travis builds would likely be interpreted by us as "all good" when in fact things weren't tested at all, if we would automatically skip.

mih:
I disagree. We propagate these failures all the way to datalad proper CI tests. This leads to all PR always failing, which implies more human costs than skipping the *two* affected tests in here. I proposed #92 

Once that is merged, we can change the title here to "Re-enable tests from singularity hub is back".

bpoldrack:
Merged #92.
To be clear: I was arguing against an auto-skip on `ConnectionError`, not against consciously skipping a particular test after being aware of the reason it fails.

vsoch:
You've just used the api to download the container here, and I need to ask you to discontinue this until I've announced that the server is ready. You have a very limited quota, across all GET requests per week, so it's very unwise to be interacting with the API until I've had a chance to update everyone with these limits and best practices. thanks!

yarikoptic:
@vsoch just pushed out 0.5.0 release of this extension which skips tests querying singularity hub, so our datalad core tests (which do downstream testing of released extensions) now should stop querying it as well

vsoch:
Great, thanks!

PR #93, opened 2019-07-12
DOC+ENH: Finalize changelog for 0.5.0 and boost version

yarikoptic:
Let's kick it out so datalad core tests do not fail etc?


kyleam:
> Let's kick it out so datalad core tests do not fail etc?

I'm fine with a release, but I don't see how it when it comes to the failing tests.  The -container run is now marked as an allowed failure, so the core tests don't fail due to the -container run.

yarikoptic:
[@vsoch asks to stop querying the hub for now](https://github.com/datalad/datalad-container/issues/91#issuecomment-511482821) so let's release so we stop while testing datalad core.

yarikoptic:
uploaded to pypi, eventually will build fresh debian pkg. Cheers!

PR #99, opened 2019-08-21
BF: use sys.executable not just python for docker adapter invocation

yarikoptic:
Debian datalad-container package provides only python3 bindings.
datalad cmdline uses python3 not python.  So trying to use
datalad containers-run fails.  By hardcoding sys.executable we have
guarantee that correct python is used


kyleam:
oops, thank you

PR #104, opened 2019-09-17
Docker adapter tests

kyleam:
Add some simple tests for the docker adapter that we can build on as we start to deal with some rough edges.  Hopefully my adjustment to .travis.yml is enough to make these run on Travis, but it might need additional tune-ups.

(I'm also sneaking in a commit that fixes calls to `logging.exception()`.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@
 # travis-ci.org definition for DataLad build
 language: python
 
+services:
+  - docker
+
 git:
   # we use submodules to be managed with datalad
   submodules: false
--- a/datalad_container/adapters/docker.py
+++ b/datalad_container/adapters/docker.py
@@ -188,9 +188,10 @@ def main(args):
 if __name__ == "__main__":
     try:
         main(sys.argv)
-    except sp.CalledProcessError as exc:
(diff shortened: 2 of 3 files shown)


yarikoptic:
sorry that I have missed this one completely. thanks for the PR

I would have only changed location for `test_docker.py` -- should go under `datalad_container/adapters/tests/test_docker.py` (so `tests/` submodule should be created, do not forget to add an `__init__.py` there, we are still py2 here).

Tests seems to run:
```
datalad_container.adapters.test_docker.TestAdapterBusyBox.test_containers_run ... ok
1540datalad_container.adapters.test_docker.TestAdapterBusyBox.test_save_and_run ... ok
1541datalad_container.adapters.test_docker.test_docker_save_doesnt_exist ... ok
```

I will look into enabling codecov for the extension(s) now (apparently it is not)

yarikoptic:
coverage -- actually it is all good/enabled.  I guess it was done after this elderly PR came to life or there was a glitch with codecov so report didn't appear here (which is what triggered me to look into it)

kyleam:
re: location  Thanks for catching that.  I'll push a fixup.

PR #109, opened 2019-10-28
Be compatible with datalad 0.12.0rc6

### LINKED ISSUE #108: RF: rev_resolve_path() -> resolve_path()
Renamed in -core. https://github.com/datalad/datalad/issues/3797

bpoldrack:
- Closes #108 
- add test for `find_container`, since it's absence is why we missed the failure in datalad's extension tests
- stumbled upon and fixed a minor bug in `find_container`: We implemented to just return the only container we found if no name was provided (and there's only one indeed), but `container_name` was a mandatory parameter, rendering that functionality pointless

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ git:
   submodules: false
 
 python:
-  - 2.7
   - 3.5
   - 3.6
 
--- a/datalad_container/find_container.py
+++ b/datalad_container/find_container.py
@@ -25,7 +25,10 @@ def _get_container_by_name(_, name, containers):
 
 def _get_container_by_path(ds, name, containers):
     from datalad.distribution.dataset import resolve_path
-    container_path = resolve_path(name, ds)
+    # Note: since datalad0.12.0rc6 resolve_path returns a Path object here,
+    #       which then fails to equal c['path'] below as this is taken from
+    #       config as a string
+    container_path = str(resolve_path(name, ds))
     container = [c for c in containers.values()
                  if c['path'] == container_path]
     if len(container) == 1:
--- a/datalad_container/find_container.py
+++ b/datalad_container/find_container.py
@@ -38,7 +38,7 @@ def _get_container_by_path(ds, name, containers):
 # Entry point
 
 
-def find_container(ds, container_name):
+def find_container(ds, container_name=None):
     """Find the container in dataset `ds` specified by `container_name`.
 
     Parameters
--- a/datalad_container/tests/test_find.py
+++ b/datalad_container/tests/test_find.py
@@ -1,0 +1,38 @@
+import os.path as op
+from datalad.api import Dataset
+from datalad.tests.utils import (
+    ok_clean_git,
+    assert_in,
+    assert_is_instance,
+    assert_in_results,
+    assert_result_count,
+    assert_raises
+)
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -7,7 +7,6 @@ git:
   submodules: false
 
 python:
-  - 2.7
```
yarikoptic: why?  aren't we testing here against released version of datalad, and testing datalad master against this in its own .travis.yml?

bpoldrack: Apparently not. Those builds failed with "datalad requires python >= 3.5"

kyleam:
> * Closes #108

Thanks.

> * add test for `find_container`, since it's absence is why we missed the failure in datalad's extension tests

OK, thanks.  It [was covered][0] at the time, but that coverage must have been lost somewhere along the way (perhaps when we started skipping the SingularityHub-dependent tests).

[0]: https://codecov.io/gh/datalad/datalad-container/compare/ac9202c36b03dfe2ca9124defb7342386c502f24...ac8bbf4ee2506961d9806b6859fe638faa5b67d1/diff

> * stumbled upon and fixed a minor bug in `find_container`: We implemented to just return the only container we found if no name was provided (and there's only one indeed), but `container_name` was a mandatory parameter, rendering that functionality pointless

I disagree that that was a bug.  When unspecified, the container name comes in to `containers_run` as None, and that's what it passes along to `find_container`.  Making it optional means that the caller _could_ drop the argument and still get the None behavior, but the only relevant caller in the code base is `containers_run`, so I don't see how it really matters.

.travis.yml
kyleam: I believe since the beginning this has tested against a datalad installed from `git+https://github.com/datalad/datalad.git`.

bpoldrack:
> (perhaps when we started skipping the SingularityHub-dependent tests).

Yes, it's implicitly covered in `containers-run` tests, that are skipped due to Singularity-Hub issue.

> but the only relevant caller in the code base is containers_run, so I don't see how it really matters.

Agree. Doesn't really matter. Just encountered it when writing the test and python complained that the argument was required. Found that a bit weird and more intuitive to then have it optional. But yes - "bug" is a bit strong a word ;-)

.travis.yml
bpoldrack: Double-checked: `requirements.txt` lets us install master.

bpoldrack:
@kyleam : Since you have a point we could still behave as if we were ;-)
I'll change it.

kyleam:
I'm going to cancel the pending build because that is just for a commit message change, and the [previous build passed][0].

re: dropping python 2 from build matrix: That's necessary given the current state of only testing against datalad's master.  This PR is the first to encounter that issue because it's the first since datalad's master dropped python 2.  We probably want to add at least one python 2 run that tests with the latest datalad release, but that's of course not for this PR to do.

So I'll merge this in.

[0]: https://travis-ci.org/datalad/datalad-container/builds/603892938

.travis.yml
yarikoptic: doh, indeed.  but I think that is a bug since none other extension (besides revolution) does it AFAIK, and it is against on what we have agreed in the past.  I will adjust #110 for that

PR #110, opened 2019-10-28
RF: Prune not needed for this extension installations for Travis, reenable PY2 testing, test against released datalad

yarikoptic:
were pertinent to datalad (sudo -E) and datalad-neuroimaging (heudiconv),
not here

In followup to merged #109 
- reverting drop of testing against PY2 
- as we agreed, extensions should test against released version of datalad, while datalad master would test itself against extensions.  This extension was the only one which had master datalad installed in requirements.txt


yarikoptic:
woohoo -- finally all legit again! ;) unless no objections, will merge soon (or feel welcome to merge)

PR #114, opened 2019-12-04
DOC: basic demo: Update stale statement about Docker support

kyleam:
When that was written, the docker adapter wasn't available.  It's now
in the main source tree and is wired up to containers-add.

Re: https://github.com/datalad-handbook/book/pull/242#discussion_r353793791


bpoldrack:
Obviously non-controversial. Thanks.

PR #117, opened 2020-02-23
Version boost (to 1.0.0) and changelog entry

yarikoptic:
any objections?


kyleam:
> any objections?

none from me

yarikoptic:
Then let it be.  I will "ff" locally and release

PR #119, opened 2020-05-04
RF: run: Filter eval_params from inherited parameters

kyleam:
```
containers-run builds ContainersRun._params_ from Run._params_ to
extend run's interface with container-specific parameters.  However,
this is unintentionally populating ContainersRun._params_ with
parameters from eval_params because build_doc() modifies Run._params_
in place.

Having eval_param keys in ContainersRun._params_ doesn't currently
cause any issues because build_doc() does a blanket update of _params_
with eval_params, but a proposed change to build_doc() exposed the
issue (datalad/datalad#4480).  That patch will likely be rewritten in
a way that doesn't trigger the issue, but, in hopes of avoiding
similar issues in the future, let's rework ContainersRun handling to
not depend on internal details of how build_doc() works.
```

Ref: datalad/datalad#4480

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -13,6 +13,7 @@ from datalad.distribution.dataset import datasetmethod
 from datalad.distribution.dataset import require_dataset
 from datalad.interface.utils import eval_results
 
+from datalad.interface.common_opts import eval_params
 from datalad.interface.results import get_status_dict
 from datalad.interface.run import Run
 from datalad.interface.run import run_command
@@ -28,7 +29,7 @@ lgr = logging.getLogger("datalad.containers.containers_run")
 CONTAINER_NAME_ENVVAR = 'DATALAD_CONTAINER_NAME'
 
 _run_params = dict(
-    Run._params_,
+    {k: v for k, v in Run._params_.items() if k not in eval_params},
     container_name=Parameter(
         args=('-n', '--container-name',),
         metavar="NAME",


yarikoptic:
Thank you @kyleam. I wouldn't mind having it fixed as proposed, at least in a short run. But
I have a concern though - this change would be to introduce use `eval_params` outside of datalad core.  Not that we aren't already using here `_params_`, and even though `eval_params` is not a protected/private attribute, it feels that we would be tie-ing into internal datalad core code assumptions here. I am not sure if anywhere else we would desire similar functionality of copying parameters from one class to another, but may be it would be time to introduce some supporting functionality (`get_class_params`?) in datalad core which would do the right thing, and then use it here, instead of directly wrangling `_params_` and `eval_params`?

kyleam:
> Not that we aren't already using here `_params_`, and even though `eval_params` is not a protected/private attribute, it feels that we would be tie-ing into internal datalad core code assumptions here.

The code as is assuming that `_params_` isn't modified by `build_doc`.  I actually view that as a stronger reliance on internal details than relying on `eval_params` existing.

> I am not sure if anywhere else we would desire similar functionality of copying parameters from one class to another, but may be it would be time to introduce some supporting functionality (`get_class_params`?) in datalad core which would do the right thing, and then use it here, instead of directly wrangling `_params_` and `eval_params`?

Either way we need to assume `_params_` exists.  It's part of the interface that extensions rely on.  So what's left is trading the assumption that `eval_params` exists and is okay to use with the assumption that a helper exists and is okay to use.  Perhaps the argument is that, when using the helper function, the caller doesn't have to be aware of the internal details that `build_doc` mutates `_params_`.  But that doesn't seem to hold much water, because the caller wouldn't use the helper and would just access `_params_` directly if they didn't know that.  I'm failing to see where we come out ahead.

Another option would be to change `build_doc` to not modify `_params_`, making the current assumption in containers-run correct.  That's a bit more intrusive and risks breakage, but it might be worth exploring.

kyleam:
> Another option would be to change `build_doc` to not modify `_params_`, making the current assumption in containers-run correct. That's a bit more intrusive and risks breakage, but it might be worth exploring.

Yet another, more conservative option would be to make `build_doc` filter out `eval_params` from `_params_`, with the understanding that interfaces might want to pull in and extend another class's `_params_`.

Hmm, that might be the way to go.

kyleam:
> Yet another, more conservative option would be to make `build_doc` filter out `eval_params` from `_params_`, [...]

Doy, on second thought, that's effectively what `build_doc` already does, as I described in the initial post.  So this is still an option, but it'd be reduced to "explicitly document this in `build_doc`s docstring, along with a comment that explains the importance of the `.update` call".

kyleam:
datalad/datalad#4482 explicitly documents the behavior that containers-run relies on.  Assuming that gets merged, this can be closed without being merged.

Issue #126, opened 2020-09-08
Recent test_schemes.test_docker error in datalad core test runs


kyleam:
An error in test_schemes.test_docker popped up in datalad PRs in the last day or two.  Here's one example:

```
datalad_container.tests.test_schemes.test_docker ... ERROR
bb -> .datalad/environments/bb/image

Versions: annexremote=1.4.3 appdirs=1.4.4 boto=2.49.0 cmd:7z=16.02 cmd:annex=7.20190819+git2-g908476a9b-1~ndall+1 cmd:bundled-git=2.20.1 cmd:git=2.20.1 cmd:system-git=2.28.0 cmd:system-ssh=7.6p1 exifread=2.3.1 humanize=2.6.0 iso8601=0.1.12 keyring=21.4.0 keyrings.alt=3.4.0 msgpack=1.0.0 mutagen=1.45.1 requests=2.24.0 tqdm=4.48.2 wrapt=1.12.1
======================================================================
Obscure filename: str=b' "\';a&b&c\xce\x94\xd0\x99\xd7\xa7\xd9\x85\xe0\xb9\x97\xe3\x81\x82 `| ' repr=' "\';a&b&cΔЙקم๗あ `| '
ERROR: datalad_container.tests.test_schemes.test_docker
Encodings: default='utf-8' filesystem='utf-8' locale.prefered='UTF-8'
----------------------------------------------------------------------
Environment: LANG='C.UTF-8' PATH='/opt/hostedtoolcache/Python/3.7.9/x64/bin:/opt/hostedtoolcache/Python/3.7.9/x64:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/usr/share/rust/.cargo/bin:/home/runner/.config/composer/vendor/bin:/home/runner/.dotnet/tools:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin' GIT_CONFIG_PARAMETERS="'init.defaultBranch=master'"
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/datalad/tests/utils.py", line 177, in _wrap_skip_if_no_network
[...]

PR #127, opened 2020-09-23
Let DataLad handle shub:// when capable

kyleam:
datalad/datalad#4816 added support for `shub://` URLs within the DataLad special remote.  This PR adjusts `containers-add` to go through that when it's available.

- [X] Drop tip commit that temporarily enables tests that use Singularity Hub.
  
  Update: That build looks fine.  The failure in both jobs is the known and unresolved issue from gh-126.

  https://travis-ci.org/github/datalad/datalad-container/builds/729717666

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -29,11 +29,23 @@ from .definitions import definitions
 
 lgr = logging.getLogger("datalad.containers.containers_add")
 
+# The DataLad special remote has built-in support for Singularity Hub URLs. Let
+# it handle shub:// URLs if it's available.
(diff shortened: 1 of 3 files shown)


yarikoptic:
tested - works.  Thank you @kyleam .

Note: if datalad is `< (not yet released) 0.14`, it will still be a temporary URL registered within git-annex for the container, not the `shub://` URL.  It will be up to a user to manually fixup (`git annex registerurl`) those URLs in annex upon upgrade from temporary urls in "web remote" to the one from `updateurl` config value for that container. Ok, let's just look into the bright future for now unless user reports somehow start to appear demanding some automation on that aspect.

PR #128, opened 2020-09-23
MNT: Drop bound method imports

kyleam:
These haven't been needed since DataLad v0.11.3.  (The current minimum
version is 0.12.)


yarikoptic:
CI fails seems to be only the unrelated and known `ERROR: datalad_container.tests.test_schemes.test_docker` so let's proceed, thanks!

PR #131, opened 2020-10-09
RF: Adjust test to accomodate for no-log behavior of `run`

mih:
Introduced by https://github.com/datalad/datalad/pull/4996

### DIFF
--- a/datalad_container/tests/test_run.py
+++ b/datalad_container/tests/test_run.py
@@ -24,6 +24,10 @@ from datalad.utils import (
     on_windows,
 )
 from datalad.support.network import get_local_file_url
+from datalad.cmd import (
+    StdOutCapture,
+    WitlessRunner,
+)
 
 
 testimg_url = 'shub://datalad/datalad-container:testhelper'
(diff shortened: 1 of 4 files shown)


yarikoptic:
Sorry, I don't get it, could you elaborate on the change in behavior that PR introduces such that outside code needs fixing?

kyleam:
>  could you elaborate on the change in behavior that PR introduces such that outside code needs fixing?

Here's my guess from datalad/datalad#4996: If I understand it correctly, the reason that `swallow_outputs` works for `Runner` and not `WitlessRunner` is that `Runner` explicitly sets the `stdout` argument of `Popen` to `sys.stdout`, which allows the `swallow_outputs` override to be effective.

datalad_container/tests/test_run.py
```diff
@@ -163,21 +167,24 @@ def test_custom_call_fmt(path, local_file):
     ds.save()  # record the effect in super-dataset
 
     # Running should work fine either withing sub or within super
-    with swallow_outputs() as cmo:
-        subds.containers_run('XXX', container_name='mycontainer')
-        assert_in('image=righthere cmd=XXX img_dspath=. name=mycontainer', cmo.out)
+    out = WitlessRunner(cwd=subds.path).run(
+        'datalad containers-run -n mycontainer XXX',
```
kyleam, line 171: To make this work worth the WitlessRunner in the release, you could convert this and the other commands to a list.

kyleam:
This PR covers the `test_custom_call_fmt` failure, but `TestAdapterBusyBox.test_containers_run` showed a similar failure in the -container run at datalad/datalad#4996.

datalad_container/tests/test_run.py
kyleam: Hmm, but WitlessRunner isn't available until 0.13 and the current minimum version for -container is 0.12.  Given that we probably want to avoid introducing new uses of the old Runner and that this is just about capturing output for the tests, one option would be to use `subprocess.run`.

yarikoptic: I think it is ok to raise the minimal required version to 0.13 since 0.12 is no longer "supported" really

kyleam: I don't feel too strongly either way (and in general am not against bumping the datalad minimum version fairly aggressively in extensions), but my personal preference would be to not bump the minimum dependency for the sake of test code alone.

yarikoptic: Related: may be I should look into establishing testing of extensions against claimed minimally supported versions... I wonder if containers is still compatible with 0.12 since we never test

yarikoptic: it seems that 0.12.1 might not be good enough: https://travis-ci.org/github/datalad/datalad-container/jobs/734406987 . Will try 0.13.0 now

kyleam: > it seems that 0.12.1 might not be good enough

Thanks for looking into it.  Those failures you saw for 0.12.1 happened when you bumped to 0.13 as well.  I haven't looked more closely to try to figure out what's going on, though.

Locally I tried with a `pip -e`-installed 0.12.0, and -container's test suite passes for me.  I also tried with a `pip install datalad==0.12.1`-installed 0.12.1, and the test suite passes.  So, at least based on the test suite, I think we're still compatible with 0.12.x.

Anyway, I'm okay with bumping the minimum version to 0.13.  I'll tweak this PR to do that, and to address the outstanding issues.

yarikoptic: yeah, my min version testing causes some other fallouts which I did not have a chance to look into yet. sorry for misleading

Issue #106, opened 2019-09-17
Consider skopeo-based adapter


kyleam:
[skopeo] "is a command line utility that performs various operations on container images and image repositories".  It has a `copy` command to "copy container images between various storage mechanisms".

Here's an example that copies a container from DockerHub to a local OCI-layout directory.

```
$ skopeo copy docker://busybox:latest oci:bb:latest
Getting image source signatures
Copying blob sha256:7c9d20b9b6cda1c58bc4f9d6c401386786f584437abbe87e58910f8a9a15386b
 742.94 KB / 742.94 KB [====================================================] 0s
Copying config sha256:8cf90cc9e23fce3bb22a95933b0f1008115828369857f09825dfb376b175f897
 575 B / 575 B [============================================================] 0s
Writing manifest to image destination
Storing signatures

$ tree --charset=ascii bb
bb
|-- blobs
|   `-- sha256
|       |-- 7c9d20b9b6cda1c58bc4f9d6c401386786f584437abbe87e58910f8a9a15386b
|       |-- 8cf90cc9e23fce3bb22a95933b0f1008115828369857f09825dfb376b175f897
|       `-- 96fed174fbe8d6aeab995eb9e7fc03a6326abbc25adb5aa598e970dfe8b32c6d
|-- index.json
`-- oci-layout

2 directories, 5 files
```

I need to look into skopeo more, but here are some things to consider:

  * The above command works without docker installed or the daemon running.

  * Like the `docker save` output, the OCI layout can be nicely stored and de-duplicated as annex content in the dataset.

  * The above layout should still work with a solution to attach URLs to individual layers (gh-98).

  * This may help with podman integration (gh-89) since IIUC podman still works from a local store of containers and we'd need a similar save/load adapter.

  * Similar to the command above, we should be able to copy from the dataset directory to the local daemon storage.

  * One of the motivations of the docker adapter is lack of singularity support on Windows, so we probably wouldn't be able to completely replace the current approach with skopeo, which I don't think has any Windows support.  And `docker load` doesn't support the OCI layout (there's a [draft PR][pr] from February),

  * skopeo would be an optional dependency.

  * skopeo isn't yet packaged for Debian.

[skopeo]: https://github.com/containers/skopeo
[pr]: https://github.com/moby/moby/pull/38738

yarikoptic:
[...]

Issue #89, opened 2019-05-27
Support 'podman'


mih:
https://podman.io

> What is Podman? Podman is a daemonless container engine for developing, managing, and running OCI Containers on your Linux System. Containers can either be run as root or in rootless mode. Simply put: `alias docker=podman`.

mih:
DEB packages are available from http://ppa.launchpad.net/projectatomic

almereyda:
I would believe the work requested here only requires little adaptation for rootless containers, and could also benefit to Docker rootless.

If volumes are mapped, for example, it will be good to use the `:z` or `:Z` mount options.

Otherwise this should also work as a drop-in replacement and also removes the limitation currently present in the foot note of:

* [7.2. Computational reproducibility with software containers — The DataLad Handbook](https://handbook.datalad.org/en/latest/basics/101-133-containersrun.html#id1)

yarikoptic:
indeed, podman is gaining popularity and even for HPC (see e.g. https://github.com/NERSC/podman-hpc although I did not look into what it does there exactly). I guess similarly to docker ideally we should look also into addressing #98 first, likely through using another mechanism than we have ATM for how we place layers on a filesystem.

PR #135, opened 2020-10-30
ENH: containers-add-dhub - add multiple images/tags/repositories from docker hub

yarikoptic:
It is the first truly "containers-" (as opposed to "container-") command intended to add multiple containers from the Docker hub organization, possibly for multiple repositories.

Since docker hub announced that retention policy is about to be changed (well, deadline now moved to mid 2021 so not as urgent), we better provide easy means to "mirror" (or backup) all (or desired subset) of docker containers within a datalad dataset.  We are planing to do that within  https://github.com/repronim/containers , which ATM contains only docker containers rebuilt into singularity containers.  But IMHO it could be of great benefit to have a helper tool or even a command so any user of datalad-container could establish a seamless backup of docker container from his/her (or just a favorite collection of) docker repository .

ATM it is possible to do that with this helper script (not a `containers-` command yet) which was initially crafted by @kyleam and then tortured into spaghetti code by me, for an "official" Docker images repo (e.g. `busybox` or `neurodebian`) which is just a shortcut to `library/<repo>` on the hub, or any other than library collection of the repositories (e.g. `repronim/`).  
It already supports multiple architecture image collections (e.g. `busybox`) and annotates architecture (if multiple archs for the tag) in the image name.  See the header of the file for more information/conventions 

If eager to try (although might want to uncomment the TEST definitions in the file) - try running in some target new dataset (probably created with `-c text2git`), e.g. `.../tools/containers_add_dhub_tags.py <(echo busybox)` or `tools/containers_add_dhub_tags.py <(echo repronim/)` .
 
TODOs
- many embedded in the code ATM, yet to be fully populated here
  -  [ ] populate with URLs to layers on docker hub
  -  [ ] establish a "default" datalad container config for a repo (for a "latest" and/or `repo:tag`) which would use latest for the default arch
  - ...
- [ ] consider to make it into a command, e.g. `containers-add-dhub`

### DIFF
--- a/tools/containers_add_dhub_tags.py
+++ b/tools/containers_add_dhub_tags.py
@@ -37,6 +37,8 @@ DHUB_ENDPOINT = "https://hub.docker.com/v2"
 # or if we just make it one repository at a time, then could become CLI options
 target_architectures = '.*'
 target_tags = '.*'
+# TODO: forget_tags = 'master' -- those for which we might not want to retain prior versions
+# or may be exclude them completely since too frequently changing etc?
 
 # TEST on busybox on just a few architectures and tags - it is tiny but has too many
 #target_architectures = '^(amd64|.*86)$'
@@ -203,6 +205,9 @@ def process_files(files):
                 # image/datalad container -- we just add container entry pointing to that
                 # one.  If there is no matching one -- we do get "latest"
                 latest_matching_tag = None
+                # NOTE: "master" is also often used to signal a moving target
+                # it might, or not, correspond to tagged release.  I guess we are just
+                # doomed to breed those
                 if target_tags_re.match('latest'):
                     matching_tags = []
                     for tag, images in tag_images.items():
@@ -229,6 +234,7 @@ def process_files(files):
                     if not target_tags_re.match(tag):
                         lgr.debug("Skipping tag %(tag)s")
                         continue
+                    multiarch = len({i['architecture'] for i in images}) > 1
                     for image in images:
                         architecture = image['architecture']
                         if not target_architectures_re.match(architecture):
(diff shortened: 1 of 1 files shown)


kyleam:
Thanks for turning my initial sketch into something more useful.

I think in terms of exploration this is good, and I don't have any objections to having this script or something like it live in `tools/`.  However, having played around with skopeo more, it's really impressive, and I think we should just focus on it going forward.  For the immediate problem at hand, there is already `skopeo-sync`, and it takes care of a large part of what this script does.  A small wrapper would still be needed to list the repos of an organization and register them with containers-add (though I'm not sure that it's a good thing to couple the local dump with registering the container).  Also, I don't think skopeo-sync supports getting all the architectures yet [*].

Here's an example to sync the neurodebian repo:

```console
$ skopeo sync --src docker --scoped --dest dir docker.io/neurodebian images/
$ tree --charset=ascii images | head
images
`-- docker.io
    `-- library
        |-- neurodebian:artful
        |   |-- 261816990a775a30f88752a13a62a52bcde56bb65e4a55f197a2e9fb9bb5920e
        |   |-- 448bb314afa553bfb1578121328bbe92d2b5ca0f411967e7a0a200f672ade92f
        |   |-- 4ccdce43d1e00fd03ac5438d39e731c16db3dfcf03c68390884b8e8c814221ca
        |   |-- 518254c3dbad5ed8bf16b404277faae75f3ba8bd5fcd69a217de42fbed22f250
        |   |-- 78ff727be57a68299558bb40b737669ca5cb9a8db948411d852ec809c14e7a1f
        |   |-- 82656eee95ad054e0aa75486e7c55b7666c26abbd9bf19373dd349f6e172ce9d
```

Setting backing up/syncing a registry aside, the two missing pieces for working with skopeo are the adapter and support for the URLs in the datalad special remote.  I think the mechanics of both these aspects are straightforward, but the trickier part about both is thinking through the design to leave room for other sources and targets.  In the case of the blob downloader, that probably just boils down to including the specific registry as part of the URL.  In the case of the adapter, the main thing I have in mind is execution with podman (gh-89, gh-106).  

I'm working on getting an initial version of the downloader and skopeo adapter set up.

[*] It looks like support for that is coming: https://github.com/containers/skopeo/pull/880/

yarikoptic:
Great, thanks @kyleam for looking into it!
I also wonder, if we bet on skope's `sync`, would it still work (`sync` only new images etc) if files are annexed (symlinks) -- would we need to keep them unlocked or unlock explicitly first (pain/inefficient).

kyleam:
> I also wonder, if we bet on skope's `sync`, would it still work (`sync` only new images etc) if files are annexed (symlinks)

Hmm, good point, but actually an update wouldn't work in general with a directory destination.  It will fail refusing to overwrite the directory.  I had assumed it would skip because I saw "Copying blob X skipped: already exists" in sync output posted to the issue tracker, but, looking at the code now, it seems like that'd only be for copying to a registry destination, not a local directory.  So, it was too soon to say sync could mostly replace this script.  We could still do a `skopeo copy --all ...`, though, rather than go through the docker adapter.

yarikoptic:
I think we better just have this PR merged to provide a tool.  All the TODOs might come later if decided to proceed (likely with the next docker hub announcement ;) )

PR #137, opened 2020-11-10
WIN/BF: When on Windows, close temporary files for further use

adswa:
I tried to put my explorations from https://github.com/datalad/datalad/issues/5138 into a PR. Running tests locally on Windows is a potpourri of errors, so pushed this hoping that the CI catches everything that I might have broken with this for other operating systems ;-)

### DIFF
--- a/datalad_container/adapters/docker.py
+++ b/datalad_container/adapters/docker.py
@@ -44,6 +44,8 @@ def save(image, path):
     # Use a temporary file because docker save (or actually tar underneath)
     # complains that stdout needs to be redirected if we use Popen and PIPE.
     with tempfile.NamedTemporaryFile() as stream:
+        # Windows can't write to an already opened file
+        stream.close()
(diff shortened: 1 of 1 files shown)


datalad_container/adapters/docker.py
```diff
@@ -44,6 +48,9 @@ def save(image, path):
     # Use a temporary file because docker save (or actually tar underneath)
     # complains that stdout needs to be redirected if we use Popen and PIPE.
     with tempfile.NamedTemporaryFile() as stream:
+        if on_windows:
```
yarikoptic, line 51: do we really need to condition it with `on_windows`? sounds like it could just be generally done, or am I missing something?

adswa: Oh, true. For some reason, I didn't think of this. Will adjust.

kyleam (approved):
Great, thanks!

kyleam:
I've cancelled one of the dup Travis builds.

Issue #123, opened 2020-07-17
cannot stop docker ran containers with md5sum waiting on stdin


yarikoptic:
and that containers/alpine is actually a 

(even can't `^Z`  it)

```shell
(git-annex)lena:~/proj/repronim/demo-study[master]git-annex
$> datalad containers-run -n containers/alpine md5sum                                                                        
[INFO   ] Making sure inputs are available (this may take some time) 
[INFO   ] == Command start (output follows) ===== 
^C^C^C^C
^C^C^C^C^C^C^C

^C^C^C

^C^C^C^C

^C^C^C^Z
```

I had to `docker stop` that container explicitly to get rid of this

<details><summary>DataLad 0.13.0.dev302 WTF (configuration, datalad, dataset, dependencies, environment, extensions, git-annex, location, metadata_extractors, python, system)</summary>

# WTF
## datalad 
  - full_version: 0.13.0.dev302-g8ed48
  - version: 0.13.0.dev302
## dataset 
  - id: c7947854-c845-11ea-8e05-87261519fb0a
  - metadata: <SENSITIVE, report disabled by configuration>
  - path: /home/yoh/proj/repronim/demo-study
  - repo: AnnexRepo
## dependencies 
  - annexremote: 1.4.3
  - appdirs: 1.4.3
  - boto: 2.49.0
  - cmd:7z: 16.02
  - cmd:annex: 8.20200617+git192-g5849bd634-1~ndall+1
  - cmd:bundled-git: 2.24.0
  - cmd:git: 2.24.0
  - cmd:system-git: 2.27.0
  - cmd:system-ssh: 8.1p1
  - exifread: 2.1.2
  - git: 3.1.0
  - gitdb: 4.0.2
  - humanize: 2.3.0
  - iso8601: 0.1.12
  - keyring: 18.0.1
  - keyrings.alt: 3.4.0
  - msgpack: 0.6.2
  - mutagen: 1.40.0
  - requests: 2.23.0
  - tqdm: 4.43.0
  - wrapt: 1.11.2
## environment 
  - GIT_PAGER: less --no-init --quit-if-one-screen
  - LANG: en_US.UTF-8
[...]

PR #142, opened 2020-11-17
BF: docker adapter: Don't couple 'run -i' with stdin.isatty

kyleam:
24af27f9 (BF: docker adapter: Drop -it flags when stdin is not a tty,
2019-11-11) dropped both --interactive and --tty from the 'docker run'
call if stdin.isatty() reports false, avoiding an "the input device is
not a TTY" failure in this case.

However, dropping --interactive shouldn't be necessary, and doing so
prevents the command from receiving stdin.

---

I noticed this issue when looking into gh-123.

Before this PR:

```console
$ echo foo | md5sum
d3b07384d113edec49eaa6238ad5ff00  -
$ echo foo | datalad containers-run -n alp md5sum
[INFO] Making sure inputs are available (this may take some time)
[INFO] == Command start (output follows) =====
d41d8cd98f00b204e9800998ecf8427e  -
[INFO] == Command exit (modification check follows) =====
[INFO] Total: starting
[INFO]
[INFO] Total: processed result for /tmp/po-TYjIvuy
[INFO] Total: done
action summary:
  get (notneeded: 1)
  save (notneeded: 1)
```

After:

```console
$ echo foo | datalad containers-run -n alp md5sum
[INFO] Making sure inputs are available (this may take some time)
[INFO] == Command start (output follows) =====
d3b07384d113edec49eaa6238ad5ff00  -
[INFO] == Command exit (modification check follows) =====
[INFO] Total: starting
[INFO]
[INFO] Total: processed result for /tmp/po-TYjIvuy
[INFO] Total: done
action summary:
  get (notneeded: 1)
  save (notneeded: 1)
```

### DIFF
--- a/datalad_container/adapters/docker.py
+++ b/datalad_container/adapters/docker.py
@@ -133,7 +133,8 @@ def cli_run(namespace):
               # should be configurable.
               "-v", "{}:/tmp".format(os.getcwd()),
               "-w", "/tmp",
-              "--rm"]
+              "--rm",
+              "--interactive"]
     if not on_windows:
         # Make it possible for the output files to be added to the
         # dataset without the user needing to manually adjust the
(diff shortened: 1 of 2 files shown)


kyleam:
I'll plan to wait another day or two for objections and then merge.

PR #145, opened 2021-01-16
RF: mock -> unittest.mock  and kick out 1.1.2 release

yarikoptic:
`datalad containers-run` crashed on OSX having installed datalad and datalad-container via conda.  `mock` is not listed among dependencies in feedstock I guess.  Instead of fixing feedstock, decided to fix right here -- we dropped use of mock in datalad core and can use unittest.mock here as well - no python2 support for awhile now.


kyleam (approved):
LGTM, thanks.

PR #148, opened 2021-02-03
TST: docker_adapter: Skip tests if 'docker pull' in setup fails

kyleam:
TestAdapterBusyBox.setup_class() calls 'docker pull' to bring in the
busybox image.  If this fails [1], it's probably due to rate limiting.
Regardless this is a plain 'docker pull' that doesn't involve any
parts of -container, so it's an indication that the system isn't
capable of carrying out these tests.

[1]: For example:
     https://github.com/datalad/datalad/pull/5404/checks?check_run_id=1824553695

### DIFF
--- a/datalad_container/adapters/tests/test_docker.py
+++ b/datalad_container/adapters/tests/test_docker.py
@@ -57,7 +57,11 @@ class TestAdapterBusyBox(object):
             cls.image_existed = True
         else:
             cls.image_existed = False
-            WitlessRunner().run(["docker", "pull", cls.image_name])
+            try:
+                WitlessRunner().run(["docker", "pull", cls.image_name])
+            except CommandError:
+                # This is probably due to rate limiting.
+                raise SkipTest("Plain `docker pull` failed; skipping")
 
     @classmethod
     def teardown_class(cls):


bpoldrack (approved):

yarikoptic:
I am just afraid that it would contribute even more to jumpiness of codecov etc.  I am still thinking on what we could do to resolve it "more properly".  Some thoughts:

Would could add a DOCKER_TOKEN secret to travis, and then [login](https://docs.docker.com/docker-hub/access-tokens/) before pull, but then it would not be in effect for PRs coming from clones, as our regular workflow in datalad land.

It could be paired with "pre pulling/loading": We could provide "saved" tarballs for those and pre-pull before running tests, and if `docker pull` fails - `wget -O- URL | docker load` to ensure local copy being present thus that `docker pull` in tests succeeding.

kyleam:
> I am just afraid that it would contribute even more to jumpiness of codecov etc.

In my view you're paying too much mind to codecov.

> I am still thinking on what we could do to resolve it "more properly".

Okay, no objections from me if you want to do that.  Feel free to merge or close this PR in the meantime.

PR #151, opened 2021-04-13
Set up workflow with auto for releasing & PyPI uploads

yarikoptic:
Learning to do it by following @jwodder 's example in https://github.com/datalad/datalad-crawler/pull/86

Here we can even have a release now (1.1.2-16-geccfdf8) although probably auto could perform suboptimally since merged PRs were not properly labeled.  Could that be done "postmortem" (label PRs after they are already merged) to the desired effect @jwodder ?

### DIFF
--- a/.autorc
+++ b/.autorc
@@ -1,0 +1,7 @@
+{
+    "onlyPublishWithReleaseLabel": true,
+    "baseBranch": "master",
+    "author": "auto <auto@nil>",
+    "noVersionPrefix": true,
+    "plugins": ["git-tag"]
+}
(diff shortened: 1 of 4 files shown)


jwodder:
@yarikoptic Yes, the PRs just need to be labelled at the moment that the "release"-labelled PR is merged.

yarikoptic:
ok, marked those two PRs.  @mih @kyleam - any objections to auto-mate releases here?  
this would allow for some "first hands" experience to possibly (re)consider it for datalad (core).

mih:
I don't mind here. Development pace is very calm and concerned with isolated fixes and features.

yarikoptic:
Ok, thanks for blessing @mih, let's proceed then!

PR #152, opened 2021-04-19
BF+RF: no need to pandoc long description for pypi + correctly boost MODULE/version.py for the release

yarikoptic:


yarikoptic:
ok, let's see

PR #154, opened 2021-05-30
BF:Docs: replace incorrect dashes with spaces in command names

### LINKED ISSUE #153: Documentation: commands contain incorrect dashes
Originally reported in https://github.com/datalad/datalad/issues/3671

The documentation for the commands contain dashes that should be spaces, ie:

`datalad-containers-run` should be `datalad containers-run`

loj:
The documentation for the commands contain dashes that should be spaces, ie:

`datalad-containers-run` should be `datalad containers-run`

This applies a fix that was made for other docs (datalad/datalad#1761).

(Closes #153)

### DIFF
--- a/formatters.py
+++ b/formatters.py
@@ -197,7 +197,7 @@ class RSTManPageFormatter(ManPageFormatter):
 
     def _mk_title(self, prog):
         # and an easy to use reference point
-        title = ".. _man_%s:\n\n" % prog
+        title = ".. _man_%s:\n\n" % prog.replace(' ', '-')
         title += "{0}".format(prog)
         title += '\n{0}\n\n'.format('=' * len(prog))
         return title
--- a/setup_support.py
+++ b/setup_support.py
@@ -100,12 +100,12 @@ class BuildManPage(Command):
             for cmdname in cmdline_command_names:
                 p = self._parser[cmdname]
                 cmdname = "{0}{1}".format(
-                    'datalad-' if cmdname != 'datalad' else '',
+                    'datalad ' if cmdname != 'datalad' else '',
                     cmdname)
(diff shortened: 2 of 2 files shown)


formatters.py
```diff
@@ -197,7 +197,7 @@ def _mk_synopsis(self, parser):
 
     def _mk_title(self, prog):
         # and an easy to use reference point
-        title = ".. _man_%s:\n\n" % prog
+        title = ".. _man_%s:\n\n" % prog.replace(' ', '-')
```
yarikoptic, line 200: duplication is evil and we fell for it -- we should get some proper datalad_buildsupport python package/module to avoid duplicating across all extensions etc... filed https://github.com/datalad/datalad/issues/5719

meanwhile -- this one would be the same as in https://github.com/datalad/datalad-extension-template/blob/master/_datalad_buildsupport/formatters.py#L204  so  -- good

setup_support.py
```diff
@@ -100,12 +100,12 @@ def run(self):
...
                 p = self._parser[cmdname]
                 cmdname = "{0}{1}".format(
-                    'datalad-' if cmdname != 'datalad' else '',
+                    'datalad ' if cmdname != 'datalad' else '',
                     cmdname)
                 format = cls(cmdname, ext_sections=sections, version=get_version())
                 formatted = format.format_man_page(p)
                 with open(opj(opath, '{0}.{1}'.format(
-                        cmdname,
+                        cmdname.replace(' ', '-'),
```
yarikoptic, line 108: the same as https://github.com/datalad/datalad-extension-template/blob/master/_datalad_buildsupport/setup.py#L154

yarikoptic:
Thank you @loj, sorry for the delay acting upon this PR

PR #164, opened 2022-02-02
Clarify documentation for docker usage

### LINKED ISSUE #159: Document usage with docker
Based on the documentation [here](http://docs.datalad.org/projects/container/en/latest/generated/man/datalad-containers-add.html#u-url-url-url), I would've assumed that the command: `datalad containers-add bar -u docker://foo/bar:0.1` would try to use docker, but in fact, it tried to use Singularity (which failed, as I am running on a Mac laptop). This led me to think that maybe datalad-container doesn't work with Docker at all , which I am now convinced is not true (but is not apparent just from reading the documentation...): Running `datalad containers-add bar -u dhub://foo/bar:0.1` does seem to use docker.

At any rate, would be nice to make this a bit clearer. In particular, under what circumstances is docker used and when is singularity used.

mih:
Fixes #159

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -118,13 +118,15 @@ class ContainersAdd(Interface):
         url=Parameter(
             args=("-u", "--url"),
             doc="""A URL (or local path) to get the container image from. If
-            the URL scheme is one recognized by Singularity, 'shub://' or
-            'docker://', the command format string will be auto-guessed when
-            [CMD: --call-fmt CMD][PY: call_fmt PY] is not specified. For the
-            scheme 'dhub://', the rest of the URL will be interpreted as the
-            argument to 'docker pull', the image will be saved to the location
-            specified by `name`, and the call format will be auto-guessed if
-            not given.""",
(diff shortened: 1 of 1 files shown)


datalad_container/containers_add.py
```diff
@@ -118,13 +118,15 @@ class ContainersAdd(Interface):
...
-            the URL scheme is one recognized by Singularity, 'shub://' or
-            'docker://', the command format string will be auto-guessed when
-            [CMD: --call-fmt CMD][PY: call_fmt PY] is not specified. For the
-            scheme 'dhub://', the rest of the URL will be interpreted as the
-            argument to 'docker pull', the image will be saved to the location
-            specified by `name`, and the call format will be auto-guessed if
-            not given.""",
+            the URL scheme is one recognized by Singularity ('shub://' or
+            'docker://'), a command format string for Singularity-based
+            execution will be auto-configrued when
```
arokem, line 123: Just spotted this typo: "configrued" => "configured"

yarikoptic: thank you @arokem!  I will send a PR fix with a few more typo fixes

PR #179, opened 2022-08-29
nose -> pytest, isort imports in tests, unify requirements-devel to correspond to the form as in core

yarikoptic:
Migration to pytest should be released to close `#1018335` in debian


yarikoptic:
well, it is green -- what else could we wish from it? ;)

PR #180, opened 2022-08-30
Reenabling tests using SingularityHub

### LINKED ISSUE #91: Re-enable tests when singularity hub is back

yarikoptic:
In effect Reverting 9e488cfa85ffc886d0c6fc9256f0e57e4c846bf1
but it was easier to do this way.

Closes #91

Also fixed up one previously skipped test for pytest  migration.

### DIFF
--- a/datalad_container/tests/test_run.py
+++ b/datalad_container/tests/test_run.py
@@ -15,7 +15,6 @@ from datalad.cmd import (
 )
 from datalad.support.network import get_local_file_url
 from datalad.tests.utils_pytest import (
-    SkipTest,
     assert_false,
     assert_in,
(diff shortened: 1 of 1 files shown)


yarikoptic:
good. Adds to coverage. Not sure if worth discussing, thus will proceed

PR #192, opened 2023-02-04
Account for move of @eval_results in datalad core

yarikoptic:
Move was done in https://github.com/datalad/datalad/pull/7170


yarikoptic:
green, no brainer, and we better upload to debian the fix

Issue #198, opened 2023-02-28
ENH: Add "container metadata extractor"


asmacdo:
datalad-metalad extension should include an extractor to retrieve metadata from a container image.

In scope: singularity
out of scope (for now) OCI containers

Desired metadata ideas:
 - base image
 - base image build date
 - operating system
 - installed software & version information (reproman retrace?)
 - Vulnerability report (CVEs)
 - Image build date
 - Image location (ie hub.docker.com, quay.io)

yarikoptic:
target use case: https://github.com/ReproNim/containers 

extractor for starters should just look at registered within dataset containers i.e. the ones listed by `containers-list` command, which is also available via python API:

```
❯ pwd
/home/yoh/proj/repronim/containers
❯ python -c 'import json,datalad.api as dl; print(json.dumps(dl.containers_list(result_renderer="disabled")))' | jq '.[].path' | head
"/home/yoh/proj/repronim/containers/scripts/tests/arg-test.simg"
"/home/yoh/proj/repronim/containers/images/bids/bids-aa--0.2.0.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-afni-proc--0.0.2.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-antscorticalthickness--2.2.0-1.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-baracus--1.1.4.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-brainiak-srm--latest.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-broccoli--1.0.1.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-cpac--1.1.0_14.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-dparsf--4.3.12.sing"
"/home/yoh/proj/repronim/containers/images/bids/bids-example--0.0.7.sing"
...
```

yarikoptic:
`singularity inspect` is a nice one:

```shell
$> singularity inspect images/neurodesk/neurodesk-vina--1.2.3.simg
{
    "org.opencontainers.image.revision": "e50c6323cd703e0c2df7aa56eb18e05757df2402",
    "maintainer": "Anaconda, Inc",
    "org.opencontainers.image.document": "https://github.com/Metaphorme/AutoDock-Vina-Docker",
    "GITHUB_SHA": "da8afcba3a56737deebcc847da204c0247a139f5",
[...]

PR #190, opened 2023-01-19
Add `--extra-inputs` to `containers-add`

nobodyinperson:
This PR adds the `datalad containers-add --extra-inputs` option to specify additional files that should be included in the run-record and thus auto-fetched upon `datalad containers-run`.

Furthermore the `--call-fmt` now also has a `{img_dir}` placeholder for easier specification of files next to the image.

I added a test for `containers-add` asserting that the correct values land in `.datalad/config`. There is currently no test for `containers-run` as I couldn't get this to work in CI. I tried a shell-script as 'container' and sourced files as 'overlays', works on my machine, weird errors in CI. 

Full test script `test/run.sh` for reference, a comparable `pytest` is shipped with this PR:

```bash
#!/bin/sh
cd $(dirname "$(readlink -f "$0")")
(chmod -R +w ds;rm -rf ds) 2>/dev/null
set -x
datalad create ds
cd ds
mkdir containers
sudo apptainer build containers/alpine.sif docker://alpine
sudo apptainer overlay create --size 64 containers/overlay.img
sudo chown $USER:$USER containers/overlay.img
sudo apptainer overlay create --size 64 containers/overlay2.img
sudo chown $USER:$USER containers/overlay2.img
datalad $@ containers-add alpine -i containers/alpine.sif
datalad $@ containers-add alpine-with-overlay \
    --call-fmt 'apptainer exec --overlay {img_dir}/overlay.img {img} {cmd}' \
    --extra-inputs containers/overlay.img \
    -i containers/alpine.sif
datalad $@ containers-add alpine-with-two-overlays \
    --call-fmt 'apptainer exec --overlay {img_dir}/overlay.img --overlay {img_dir}/overlay2.img:ro {img} {cmd}' \
    --extra-inputs containers/overlay.img containers/overlay2.img \
    -i containers/alpine.sif
datalad save
echo 1 > file.txt
datalad save
cat .datalad/config
datalad $@ containers-run -n alpine -o alpine-uname-a.txt 'uname -a > {outputs[0]}'
cat alpine-uname-a.txt
datalad $@ containers-run -n alpine-with-overlay -o alpine-with-overlay-uname-a.txt 'uname -a > {outputs[0]}'
cat alpine-with-overlay-uname-a.txt
datalad $@ containers-run -n alpine-with-two-overlays -o alpine-with-two-overlays-uname-a.txt 'uname -a > {outputs[0]}'
cat alpine-with-two-overlays-uname-a.txt
```

closes #189

PR #200, opened 2023-03-07
Add singularity inspect metalad extractor

asmacdo:
Fixes: https://github.com/datalad/datalad-container/issues/198

This PR is the result of Datalad Metalad hackathon last week!

TODOs
- [x] add some rudimentary unittest
- [x] make CI happy


changelog.d/20230307_164111_austin_add_metalad_extractor.md
```diff
@@ -0,0 +1,5 @@
+### 🚀 Enhancements and New Features
+
+- Add metalad extractor using `singularity inspect`.
+  Fixes https://github.com/datalad/datalad-container/issues/198 via
+  https://github.com/datalad/datalad-next/pull/200 (by @asmacdo )
```
asmacdo, line 5:
```suggestion
 https://github.com/datalad/datalad-container/pull/200 (by @asmacdo )
```

yarikoptic (changes requested):
Also will need a rudimentary unittest on extracting metadata from a sample container. There is already some which use singularity etc, so should be similar I guess.  Test should skip if no metadata or singularity available.  But metalad should be installed in CI to actually do run that test.

requirements-devel.txt
```diff
@@ -1,2 +1,3 @@
 # requirements for a development environment
 -e .[devel]
+scriv
```
yarikoptic, line 3: this inspired me for https://github.com/datalad/datalad-extension-template/pull/56 . Then here it would be `-e .[devel,devel-utils]` if so really needed. But since this has nothing to do with adding metadata extractor -- shouldn't even be in this PR. Feel welcome to submit a separate PR with such changes.

yarikoptic:
re appveyor -- I did that fix in https://github.com/datalad/datalad-container/pull/190 already which I merged today. Just rebase your branch on current master and it should be gone.

datalad_container/extractors/metalad_container.py
```diff
@@ -0,0 +1,86 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
```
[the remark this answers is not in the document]
asmacdo: IIUC it wasn't necessary (it worked) because it was being treated as an [Implicit Namespace](https://peps.python.org/pep-0420/)

But since that wasn't deliberate, I've added an __init__.py

datalad_container/extractors/metalad_container.py
```diff
@@ -0,0 +1,86 @@
...
+from uuid import UUID
+
+CURRENT_VERSION = "0.0.1"
+
+
+try:
+    from datalad_metalad.extractors.base import DataOutputCategory, ExtractorResult, FileMetadataExtractor
+    from datalad_metalad import get_file_id
+except Exception as e:
+    raise
```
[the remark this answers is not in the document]
asmacdo: this is a placeholder for https://github.com/datalad/datalad-container/pull/200/files#r1129521291

datalad_container/extractors/metalad_container.py
```diff
@@ -0,0 +1,86 @@
...
+            })
+
+    def _singularity_inspect(self, path) -> str:
+        data = subprocess.run(
+            ["singularity", "inspect", "--json", path],
+            check=True,
+            stdout=subprocess.PIPE).stdout.decode()
+        return json.loads(data)
+
+    def _singularity_version(self) -> str:
```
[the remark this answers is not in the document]
asmacdo: Agreed its low risk, but IMO its also low reward. `singularity version` seems basically free in comparison to `inspect`?

setup.cfg
```diff
@@ -41,6 +41,9 @@ datalad.extensions =
     # valid datalad interface specification (see demo in this extensions)
     container = datalad_container:command_suite
 
+datalad.metadata.extractors =
+    singularity_inspect = datalad_container.extractors.metalad_container:MetaladSingularityInspect
```
asmacdo, line 45: I changed this to `singularity_inspect` since `metalad` seemed well implied.

datalad_container/extractors/metalad_container.py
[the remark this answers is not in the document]
asmacdo: Given that there is one instance of the class per file, (bc parent is FileMetadataExtractor), I don't think we can use the `@property` pattern. 

Looked over the [External Versions]( https://github.com/datalad/datalad/blob/8d431b654337202bd6856c08240c22e370b410dd/datalad/support/external_versions.py#L153-L164) Cool, we have something very similar in ReproMan (...probably exactly this  :) This is a cool pattern, but IMO not worth the time for this particular performance bump.

However, this does seem like a cool and reasonable feature request for metalads `BaseMetadataExtractor`. I'm thinking `def _get_cacheable_metadata`. Shall I file an issue of there?

setup.cfg
```diff
@@ -41,6 +41,9 @@ datalad.extensions =
     # valid datalad interface specification (see demo in this extensions)
     container = datalad_container:command_suite
 
+datalad.metadata.extractors =
+    metalad_container = datalad_container.extractors.metalad_container:MetaladContainer
```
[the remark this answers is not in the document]
asmacdo: Easier than expected because this endpoint is under `datalad.metalad`

Without `metalad` installed:

``` 
datalad meta-extract -d . singularity_inspect images/bids/bids-pymvpa--1.0.2.sing  | jq
datalad: Unknown command 'meta-extract'.  See 'datalad --help'.

Did you mean any of these?
        metadata
```

With `metalad` installed.

```
{
  "type": "file",
  "dataset_id": "b02e63c2-62c1-11e9-82b0-52540040489c",
  "dataset_version": "9ed0a39406e518f0309bb665a99b64dec719fb08",
  "path": "images/bids/bids-pymvpa--1.0.2.sing",
  "extractor_name": "singularity_inspect",
  "extractor_version": "0.0.1",
  "extraction_parameter": {},
  "extraction_time": 1678382512.413365,
  "agent_name": "Austin Macdonald",
  "agent_email": "austin@dartmouth.edu",
  "extracted_metadata": {
    "@id": "datalad:SHA1-s993116191--cc7ac6e6a31e9ac131035a88f699dfcca785b844",
    "type": "file",
    "path": "images/bids/bids-pymvpa--1.0.2.sing",
    "content_byte_size": 0,
    "comment": "SingularityInspect extractor executed at 1678382512.2972443",
    "singularity_version": "apptainer version 1.1.5-1.fc37",
    "singularity_inspect": {
      "data": {
        "attributes": {
          "labels": {
            "org.label-schema.build-date": "Thu,_19_Dec_2019_14:58:41_+0000",
            "org.label-schema.build-size": "2442MB",
            "org.label-schema.schema-version": "1.0",
            "org.label-schema.usage.singularity.deffile": "Singularity.bids-pymvpa--1.0.2",
            "org.label-schema.usage.singularity.deffile.bootstrap": "docker",
            "org.label-schema.usage.singularity.deffile.from": "bids/pymvpa:v1.0.2",
            "org.label-schema.usage.singularity.version": "2.5.2-feature-squashbuild-secbuild-2.5.6e68f9725"
          }
        }
      },
      "type": "container"
    }
  }
}
```

datalad_container/extractors/metalad_container.py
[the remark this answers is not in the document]
yarikoptic: sorry -- this no longer leads anywhere specifically seems to me. So what is it for? may be just a comment instead of try/except would suffice for a placholder?

setup.cfg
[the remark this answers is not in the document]
yarikoptic: coolio then

datalad_container/extractors/metalad_container.py
```diff
@@ -0,0 +1,85 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+#
+#   See COPYING file distributed along with the datalad package for the
+#   copyright and license terms.
+#
+# ## ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ##
+"""MetadataRecord extractor for files stored in Datalad's own core storage"""
```
yarikoptic, line 9: adjust docstring and filename also to reflect that it is not a generic  extractor for any container

datalad_container/extractors/metalad_container.py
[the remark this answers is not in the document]
yarikoptic: the easiest way to implement is just `@functools.lru_cache(1)`  for the function, no need for any special handling/code IMHO.

datalad_container/extractors/metalad_container.py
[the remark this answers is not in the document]
asmacdo: The try/except was where I was planning on gating the plugin in case metalad isn't installed. 

Since it only has the 1 entry point, and if metalad is not installed the entry point fails before we get to this code, I dont think we need to do anything fancy with the imports.

yarikoptic:
re test
- let's use image from `tests/test_run.py:testimg_url = 'shub://datalad/datalad-container:testhelper'`
- create a session scope fixture. More advanced example is in dandi-cli https://github.com/dandi/dandi-cli/blob/HEAD/dandi/tests/fixtures.py#L342 but in your case it would be simpler -- it would `singularity pull` that image into temporary location, and give that path as what it yields as a "fixture" to underlying test. Then multiple tests could reuse that download.
- for running external commands use DataLad's `WitlessRunner` , smth like `out = WitlessRunner().run(cmd, protocol=StdOutCapture)` but see other uses in the tests of datalad.

yarikoptic:
re external_versions -- just "patch" it to provide function to load version for singularity and apptainer. Here is how

```
In [8]: external_versions.CUSTOM['cmd:singularity'] = lambda : '0.1.1'

In [9]: external_versions['cmd:singularity']
Out[9]: LooseVersion ('0.1.1')
```
so replace lambda with yours. For that probably just do in  `from datalad.support.external_versions import external_versions` and do that patching. This would make `external_versions["cmd:singularity"]` available throughout for all uses of `external_versions` in that process, including I think reporting in `datalad wtf` output as long as datalad_container is imported (via `datalad.api` or elsehow)

asmacdo:
[  ] skipif test if datalad metalad not installed
[ ] install metalad for travis

datalad_container/extractors/metalad_container.py
```diff
@@ -0,0 +1,85 @@
+# emacs: -*- mode: python; py-indent-offset: 4; tab-width: 4; indent-tabs-mode: nil -*-
+# ex: set sts=4 ts=4 sw=4 noet:
```
asmacdo, line 2: @yarikoptic is there a reason we set noexpand tab? Seems wrong for python.

yarikoptic: yeah, I think this some historical artifact. Feel welcome to fix everywhere (in separate PR) in all extensions . I think we did similar in datalad core repo but even there not everywhere

```shell
❯ git grep -h '^# ex:' | sort | uniq -c
    264 # ex: set sts=4 ts=4 sw=4 et:
     14 # ex: set sts=4 ts=4 sw=4 noet:
```

so here is a little good menial contribution to make ;)

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,62 @@
...
+            'org.label-schema.build-size',
+            'org.label-schema.usage.singularity.deffile',
+            'org.label-schema.usage.singularity.deffile.from',
+            'org.label-schema.usage.singularity.deffile.bootstrap',
+            'org.label-schema.usage.singularity.version',
+        ]
+        for label in assert_in_labels:
+            assert label in assert_in_labels
+        assert labels['org.label-schema.schema-version'] == '1.0'
+        assert result['type'] == 'container'
```
yarikoptic, line 62: add a test which actually does full metadata extraction invocation -- look how metalad does in https://github.com/yarikoptic/datalad-metalad/blob/HEAD/datalad_metalad/tests/test_extract.py testing of `meta_extract`. Also since we test on specific image -- test more - e.g. specific values (size, date), may be just a full value 
```
    "container_inspect": {
        "data": {
          "attributes": {
            "labels": {
              "org.label-schema.build-date": "Thu,_19_Dec_2019_14:58:41_+0000",
              "org.label-schema.build-size": "2442MB",
              "org.label-schema.schema-version": "1.0",
              "org.label-schema.usage.singularity.deffile": "Singularity.bids-pymvpa--1.0.2",
              "org.label-schema.usage.singularity.deffile.bootstrap": "docker",
              "org.label-schema.usage.singularity.deffile.from": "bids/pymvpa:v1.0.2",
              "org.label-schema.usage.singularity.version": "2.5.2-feature-squashbuild-secbuild-2.5.6e68f9725"
            }
          }
        },
        "type": "container"
      }
```

yarikoptic (changes requested):

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,62 @@
...
+    raise SkipTest("skipping metalad tests")
+
+# Must come after skiptest or imports will not work
+from datalad_container.extractors.metalad_container import MetaladSingularityInspect
+
+
+class TestMetaladSingularityInspect:
+
+    @with_tempfile
+    def test__singularity_inspect_nofile(self, path=None):
```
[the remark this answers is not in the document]
yarikoptic: ah, ok, makes sense - so I am ok now with it as is

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,62 @@
...
+)
+
+if not external_versions["datalad_metalad"]:
+    raise SkipTest("skipping metalad tests")
+
+# Must come after skiptest or imports will not work
+from datalad_container.extractors.metalad_container import MetaladSingularityInspect
+
+
+class TestMetaladSingularityInspect:
```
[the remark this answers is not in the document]
asmacdo: :+1: 

FYI: I put it in a class to think ahead, assuming that the container extractor will have more functionality someday, also tested in this file. The organization I've used in the past is a putting all of the tests for a function (ie singularity inspect) inside a class, and having a separate class for each function. Given that this is it for now, this could be premature.

datalad_container/extractors/_load_singularity_versions.py
```diff
@@ -0,0 +1,47 @@
...
+from datalad.support.external_versions import external_versions
+
+
+def __get_apptainer_version():
+    try:
+        out = WitlessRunner().run("apptainer --version", protocol=StdOutCapture)
+        version = out['stdout'].strip()
+
+    except FileNotFoundError as e:
+        return None
```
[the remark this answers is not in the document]
asmacdo: I was able to trim this up nicely, but not able to fully go with this suggestion, I still have to strip "apptainer " from the string, and for singularity, I have to make sure that apptainer version is not the same.

datalad_container/extractors/_load_singularity_versions.py
```diff
@@ -0,0 +1,35 @@
...
+    StdOutCapture,
+    WitlessRunner,
+)
+from datalad.support.external_versions import external_versions
+
+
+def __get_apptainer_version():
+    version = WitlessRunner().run("apptainer --version", protocol=StdOutCapture)['stdout'].strip()
+    strip_str = len("apptainer version ")
+    return version[strip_str:]
```
yarikoptic, line 20: another pattern for the same thing for your consideration

```suggestion
    return version.split("apptainer version", 1)[1].strip()
```

datalad_container/extractors/_load_singularity_versions.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def __get_singularity_version():
+    version = WitlessRunner().run("singularity version", protocol=StdOutCapture)['stdout'].strip()
+
+    # It may be possible to have both apptainer and singularity installed.
+    # If singularity is installed independently, the versions will not match.
```
yarikoptic, line 27: is that guaranteed -- they really cannot be the same?

asmacdo: Now that you mention it, I think it's ok to have cmd: singularity return the apptainer version.

It wouldn't be ideal as the code currently exists because it collects both singularity and apptainer version info, but once I use the dynamic command as you suggested, it's a moot point.

+1 I'm convinced

changelog.d/20230307_164111_austin_add_metalad_extractor.md
```diff
@@ -0,0 +1,5 @@
+### 🚀 Enhancements and New Features
+
+- Add metalad extractor using `singularity inspect`.
```
yarikoptic, line 3: not liking name 100% but due to singlularity/apptainer duality let's keep as is and do mention in changelog
```suggestion
- Add metalad extractor `container_inspect` using `singularity` or `apptainer` `inspect`.
```

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,75 @@
...
+                    'org.label-schema.usage.singularity.deffile.from': 'debian:stable-slim', 'org.label-schema.usage.singularity.version':
+                    '2.5.0-feature-squashbuild-secbuild-2.5.0.gddf62fb5'
+                }
+            }
+        },
+        'type': 'container'
+    }
+    assert result == expected_result
+
+def test_extract(singularity_test_image):
```
yarikoptic, line 67: should be skipped if no singularity or apptainer installed.

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,75 @@
...
+)
+
+if not external_versions["datalad_metalad"]:
+    raise SkipTest("skipping metalad tests")
+
+# Must come after skiptest or imports will not work
+from datalad_container.extractors.metalad_container import MetaladContainerInspect
+
+
+@with_tempfile
```
yarikoptic, line 37: next two tests rely on `apptainer` to be present . ATM AFAIK our CI installs singularity... could we make it more flexible -- e.g. add some module level const "CONTAINER_SYSTEM", skip entire module if None available, and store  string either `"singularity"` or `"apptainer"` and then use it in this tests, thus avoiding hardcoding any and allowing for either to be used here.

that would also avoid possible shadowing of CalledProcessError due to simply not finding that executable instead of it unable to work with  bad path

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,75 @@
...
+def test__container_inspect_nofile(path=None):
+    """Singularity causes CalledProcessErorr if path DNE."""
```
yarikoptic, line 39:
```suggestion
    """Singularity causes CalledProcessError if path DNE."""
```

yarikoptic (changes requested):
The main point is to make tests "execute" (not just skip) and pass on CI

datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -0,0 +1,82 @@
...
+    }
+    assert result == expected_result
+
+def test_extract(singularity_test_image):
+    ds = singularity_test_image["ds"]
+    path = singularity_test_image["img_path"]
+    result = meta_extract(dataset=ds, extractorname="container_inspect", path=path)
+    assert len(result) == 1
+
+    assert "extracted_metadata" in result[0]["metadata_record"]
```
yarikoptic, line 80:
```suggestion
    assert  result[0]["metadata_record"]["extracted_metadata"]
```
so we know that at least some content is assigned and not only key defined

datalad_container/tests/fixtures/singularity_image.py
```diff
@@ -0,0 +1,29 @@
...
+        url=TEST_IMG_URL,
+        image=fixture_file_name,
+    )
+    img_path = ds.pathobj / fixture_file_name
+    ds.get(img_path)
+    return {"ds": ds, "img_path": img_path}
+
+@pytest.fixture(scope="session")
+def container_command():
+    """Not a very useful function other than to add session scope."""
```
yarikoptic, line 28: could have just made that function "lazy" via using e.g. `@functools.cache` but ok for now

PR #204, opened 2023-03-14
add procedure to configure non-annexing of docker metadata

### LINKED ISSUE #197: gitattributes to not annex docker metadata
I wonder if datalad-container should edit `.datalad/.gitattributes` when adding docker images to include
```
environments/**/*.json annex.largefiles=nothing
environments/*/image/repositories annex.largefiles=nothing
```
The json files are really small and better handled by git.
It would also avoid errors where the layers are already in the local docker service and can avoid pulling them from the repo, but still need the manifest to check for hashes.

bpinsard:
resolves #197

### DIFF
--- a/datalad_container/resources/procedures/cfg_containers.py
+++ b/datalad_container/resources/procedures/cfg_containers.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+"""Procedure to configure dataset for containers
+"""
+
+import sys
+
+from datalad.distribution.dataset import require_dataset
+from datalad.support import path as op
+
+ds = require_dataset(
+    sys.argv[1],
+    check_installed=True,
+    purpose='Containers dataset configuration')
+
+# unless taken care of by the template already, each item in here
+# will get its own .gitattributes entry to keep it out of the annex
+# give relative path to dataset root (use platform notation)
+force_in_git = [
+    op.join('environments','**','*.json'),
+    op.join('environments','*','image','repositories'),
+]
+# make an attempt to discover the prospective change in .gitattributes
+# to decide what needs to be done, and make this procedure idempotent
+# (for simple cases)
+attr_fpath = op.join(ds.path, '.datalad', '.gitattributes')
+if op.lexists(attr_fpath):
+    with open(attr_fpath, 'rb') as f:
+        attrs = f.read().decode()
+else:
+    attrs = ''
+
+# amend gitattributes, if needed
+ds.repo.set_gitattributes([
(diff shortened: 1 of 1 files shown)


datalad_container/resources/procedures/cfg_containers.py
```diff
@@ -0,0 +1,44 @@
...
+    sys.argv[1],
+    check_installed=True,
+    purpose='Containers dataset configuration')
+
+# unless taken care of by the template already, each item in here
+# will get its own .gitattributes entry to keep it out of the annex
+# give relative path to dataset root (use platform notation)
+force_in_git = [
+    op.join('environments','**','*.json'),
+    op.join('environments','*','image','repositories'),
```
yarikoptic, line 20: I don't think we want it OS specific here, probably should just be posix, so
```suggestion
    'environments/**/*.json',
    'environments/*/image/repositories',
```

yarikoptic (changes requested):
I wonder if `cfg_containers ` wouldn't be too broad/non-specific, this specific to docker manifests right? so smth more like `cfg_dockermanifests2git` (I know - mouth full).

But then I wonder why I didn't ask why you just don't use generic `cfg_text2git`?  may be that would obliterate the need for the PR as well?

and as with anything else we would need some basic unittest, can see e.g. https://github.com/datalad/datalad/blob/HEAD/datalad/local/tests/test_run_procedure.py#L350 for an example

yarikoptic:
I will move it to Draft. I guess we should decide first on how/if to proceed with this, or just rely on cfg_text2git may be.

adswa:
I think the issue this PR fixes is quite important to fix. Is there anything I can do to help?

adswa:
> But then I wonder why I didn't ask why you just don't use generic cfg_text2git? may be that would obliterate the need for the PR as well?

I would say that ``cfg_text2git`` goes too far. It alters much more dataset behavior than would be required to hangle containers. We quite often see cases where text2git configurations cause trouble when used without consideration for the specific usecase, recent examples were for example behavioral logfiles that were supposed to stay private (thus annexed), or a case in which gigantic textfiles were generated and placed to Git which eventually caused Git to revolt. 

> I wonder if cfg_containers wouldn't be too broad/non-specific, this specific to docker manifests right? so smth more like cfg_dockermanifests2git (I know - mouth full).

I agree that it would be useful to extend this configuration to also work with singularity images (which would be placed in ``.datalad/environments/<name>/image``. I'm not too familiar with the docker setup, but if there is a glob that could match both of them, I'd be in favor. Then the name ``cfg_containers`` would also be very fitting IMHO

Issue #199, opened 2023-02-28
docker: do not get layers in present in local docker service


bpinsard:
If ones repetitively clones datasets that contains a docker image on the same machine, it will need to get the layers from the remote while it could already be pulled in the local docker. 
For example, for parallel processing with fMRIPrep, one will clone the dataset as temp clone, but will need to get the docker layers of the fMRIPrep image, even if these are already present in the docker instance.

For now the `datalad containers-run` will run the image if present in docker but still tries fetching the layer, doing unnecessary costly get operations. For example, if the layers were dropped recklessly, it tries to get the layer, fails, but runs the container anyway. 

```
$ datalad  containers-run -n alpine
[INFO   ] Making sure inputs are available (this may take some time) 
get(error): .datalad/environments/alpine/image/31f5b0c484b3651f7c883d1be2ef442c2da71d103bc7ab20cd0b1f825e67e4e7/layer.tar (file) [not available; (Note that these git remotes have annex-ignore set: origin)]
get(impossible): .datalad/environments/alpine/image (directory) [could not get some content in /home/basile/data/tests/dataset_test/.datalad/environments/alpine/image ['/home/basile/data/tests/dataset_test/.datalad/environments/alpine/image/31f5b0c484b3651f7c883d1be2ef442c2da71d103bc7ab20cd0b1f825e67e4e7/layer.tar']]
[INFO   ] == Command start (output follows) ===== 
/tmp $ 
/tmp $ 
[INFO   ] == Command exit (modification check follows) ===== 
run(ok): /home/basile/data/tests/dataset_test (dataset) [python3 -m datalad_container.adapters.do...]
action summary:                                                                                                                                           
  get (error: 1, impossible: 1)
  run (ok: 1)
  save (notneeded: 1)
```  

I guess it would require to change 

https://github.com/datalad/datalad-container/blob/66941a9414973d60926fef42684c4a2282c988f3/datalad_container/containers_run.py#L141

I can see 2 options:
- delegate the get of the image to the docker adapter, which would first check if the image is present in the local docker.
- add a cmdline flag to not add the image as an input to the run call (at the risks of failure).

On a side note, would it make sense to add the local docker service where the dataset is installed as a special remote.

yarikoptic:
[...]

PR #205, opened 2023-03-14
Add --assume-ready for image and extra-inputs

bpinsard:
Add `--assume-ready` option to be passed to `run`, with extra options for `image` and `extra-inputs`. 
see #199 . 

TODOs
- [ ] add basic testing for added option/functionality

### DIFF
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -186,11 +186,14 @@ class ContainersRun(Interface):
                             exc,
                             ", ".join(xi_kwargs)))
                     return
-        assume_ready = [ar for ar in assume_ready if ar != 'extra-inputs']
+        else:
+            # filter the whole list as .remove only removes first instance
+            assume_ready = [ar for ar in assume_ready if ar != 'extra-inputs']
 
-        if not "image" in assume_ready:
+        if "image" not in assume_ready:
             extra_inputs.append(image_path)
-        assume_ready = [ar for ar in assume_ready if ar != 'image']
+        else:
+            assume_ready = [ar for ar in assume_ready if ar != 'image']
 
         lgr.debug("extra_inputs = %r", extra_inputs)
 
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -197,12 +197,14 @@ class ContainersRun(Interface):
 
         lgr.debug("extra_inputs = %r", extra_inputs)
 
-        if len(assume_ready) == 0:
+        if not assume_ready:
             assume_ready = None
         elif len(assume_ready) == 1:
             assume_ready = assume_ready[0]
         elif  "inputs" in assume_ready and "outputs" in assume_ready:
             assume_ready = "both"
+        else:
+            raise ValueError(f"Ended up with assume_ready={assume_ready!r} which must have not happened")
 
         with patch.dict('os.environ',
                         {CONTAINER_NAME_ENVVAR: container['name']}):
--- a/datalad_container/containers_run.py
+++ b/datalad_container/containers_run.py
@@ -194,9 +194,12 @@ class ContainersRun(Interface):
 
         lgr.debug("extra_inputs = %r", extra_inputs)
 
-        # we can assume "both" with params choices and filtering above
-        if len(assume_ready) > 1:
-            assume_ready='both'
+        if len(assume_ready) == 0:
+            assume_ready = None
+        elif len(assume_ready) == 1:
+            assume_ready = assume_ready[0]
+        elif  "inputs" in assume_ready and "outputs" in assume_ready:
+            assume_ready = "both"
 
         with patch.dict('os.environ',
                         {CONTAINER_NAME_ENVVAR: container['name']}):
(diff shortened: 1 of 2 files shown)


datalad_container/containers_run.py
```diff
@@ -133,39 +144,51 @@ def __call__(cmd, container_name=None, dataset=None,
...
+
+        if not "image" in assume_ready:
+            extra_inputs.append(image_path)
+        assume_ready = [ar for ar in assume_ready if ar != 'image']
 
         lgr.debug("extra_inputs = %r", extra_inputs)
 
+        # we can assume "both" with params choices and filtering above
+        if len(assume_ready) > 1:
+            assume_ready='both'
```
yarikoptic, line 178: Thanks for taking a stab at it! Just having a quick initial peek at it.

we will need some test for it, may be based on already existing one(s).

Note here that it should be a list I think, 
```suggestion
            assume_ready = ['both']
```
but I don't like such an implicit check/mapping... 

```shell
❯ python3 -c 'import this' | grep -e Zen -e Ex
The Zen of Python, by Tim Peters
Explicit is better than implicit.
```

yarikoptic (changes requested):
just initial peek.

datalad_container/containers_run.py
bpinsard: Agreed, made it more explicit, also I think `datalad.core.local.run._prep_worktree` takes a string ( inputs/outputs/both ) rather than a list, so I added a bit of logic to ensure that value.

datalad_container/containers_run.py
```diff
@@ -133,39 +144,54 @@ def __call__(cmd, container_name=None, dataset=None,
...
+                        'run',
+                        ds=ds,
+                        status='error',
+                        message=(
+                            'Unrecognized extra_input placeholder: %s. '
+                            'See containers-add for information on known ones: %s',
+                            exc,
+                            ", ".join(xi_kwargs)))
+                    return
+        assume_ready = [ar for ar in assume_ready if ar != 'extra-inputs']
```
yarikoptic, line 168: just move into `else` (ie when assuming them ready) and do `.remove), ie. I think

```suggestion
        else:
            assume_ready.remove('extra-inputs')
```
which makes it more specific, avoids unnecessary transformation (nothing to do if not there) etc

datalad_container/containers_run.py
```diff
@@ -133,39 +144,54 @@ def __call__(cmd, container_name=None, dataset=None,
...
+
+        if not "image" in assume_ready:
+            extra_inputs.append(image_path)
+        assume_ready = [ar for ar in assume_ready if ar != 'image']
```
yarikoptic, line 172:
```suggestion
        if "image" not in assume_ready:
            extra_inputs.append(image_path)
        else:
            assume_ready.remove('image')
```

datalad_container/containers_run.py
```diff
@@ -133,39 +144,54 @@ def __call__(cmd, container_name=None, dataset=None,
...
 
         lgr.debug("extra_inputs = %r", extra_inputs)
 
+        if len(assume_ready) == 0:
+            assume_ready = None
+        elif len(assume_ready) == 1:
+            assume_ready = assume_ready[0]
+        elif  "inputs" in assume_ready and "outputs" in assume_ready:
+            assume_ready = "both"
```
yarikoptic, line 181: little of generic coding + proofing:

```suggestion
        if not assume_ready:
            assume_ready = None
        elif len(assume_ready) == 1:
            assume_ready = assume_ready[0]
        elif  "inputs" in assume_ready and "outputs" in assume_ready:
            assume_ready = "both"
        else:
            raise ValueError(f"Ended up with assume_ready={assume_ready!r} which must have not happened")
```

yarikoptic (changes requested):
Left some minor recommendations but overall I think it should be ok. 
But now it would need some basic test added since code with tests is ... broken code ;)  See e.g. recently added tests for `extra_inputs` handling in https://github.com/datalad/datalad-container/blob/HEAD/datalad_container/tests/test_run.py#L195

datalad_container/containers_run.py
yarikoptic: good point -- I like when code is more robust.  But place it under `else` to make it more explicit on when to do that and may be add a comment on why not `.remove` since it was not obvious to old me today and might raise question tomorrow when I look at the code again.

yarikoptic:
I will move it to Draft. Please bring it back to "Ready for Review" whenever it gets extended with a test @bpinsard .

yarikoptic:
ping @bpinsard any time to finish this one up?

datalad_container/containers_run.py
```diff
@@ -39,11 +40,21 @@
...
+        constraints=EnsureChoice(None, "image", "inputs", "outputs", "extra-inputs"),
+        doc="""Assume that inputs do not need to be retrieved and/or outputs do not
+        need to unlocked or removed, or containers/extra-inputs to be retrieved,
+        before running the command. This option allows
+        you to avoid the expense of these preparation steps if you know that they
+        are unnecessary."""),
 )
 
 
+
```
yarikoptic, line 57: nit: unnecessary - there is already 2 spaces above

```suggestion
```

yarikoptic:
Speed by merged
- https://github.com/datalad/datalad-container/pull/286

PR #209, opened 2023-06-07
BF: by default stop containers-run on error, to not proceed to save

yarikoptic:
containers-run parallels run. run stops on error to not save the results. I was surprised to find that failed containers-run did fail BUT also saved the results.  IMHO such dichotomy in behavior makes little sense and hence here we make on_failure to be 'error' for the containers-run.

Note: interestingly it is only the  'run' command which has any non-default default on_failure:

	❯ git grep '^ *on_failure *=[^,)(]*$'
	datalad/core/local/run.py:    on_failure = 'stop'

so all commands by default 'continue' upon failure.

### DIFF
--- a/changelog.d/pr-209.md
+++ b/changelog.d/pr-209.md
@@ -1,0 +1,3 @@
+### 🐛 Bug Fixes
+
+- BF: by default stop containers-run on error, to not proceed to save.  [PR #209](https://github.com/datalad/datalad-container/pull/209) (by [@yarikoptic](https://github.com/yarikoptic))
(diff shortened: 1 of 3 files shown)


mih (commented):
Makes sense. It would be nice to accompany any behavior (change) related updates with user-facing documentation. Over time we could build up adequate documentation, better than the sketchy notes of today. Thx!

yarikoptic:
CHANGELOG is user facing and I consider prior behavior to be a bug so for this one I will just proceed but indeed a sentence could be added to handbook at some point.

mih:
I believe no user should have to read the changelog of the entire software package or a 600 page handbook to learn about the behavior of a specific command at a specific version. The command help is the optimal place for such descriptions IMHO.

PR #210, opened 2023-06-07
Remove travis CI

yarikoptic:
I think Travis is dying off and likely I would just stop paying for it its hefty monthly price tag.

This PR is first of all check if coverage with appveyor alone would remain the same.


adswa:
How unhappy are you with the ~3.5% decrease in coverage, @yarikoptic?

yarikoptic:
gimme a few days -- I think it is worth exploring why that happened.

yarikoptic:
ok, pre-removing of `busybox:latest` image brought diff down to 1.16%

PR #212, opened 2023-06-21
Make install-singularity.sh fail if any step fails

jwodder:


jwodder:
@yarikoptic @mih @bpoldrack (Who else should I ping?) `apt-get update` is currently failing on Appveyor with the message "E: Repository 'http://ppa.launchpad.net/rabbitmq/rabbitmq-erlang/ubuntu focal InRelease' changed its 'Label' value from 'Modern Erlang packages for Ubuntu and Debian' to 'Latest (26.x) Erlang packages for Ubuntu and Debian'".  While this can apparently be worked around by passing `--allow-releaseinfo-change` to `apt-get update`, I'm not sure if this is the right thing to do in this case.

adswa:
@christian-monch did it like this in https://github.com/datalad/datalad-next/pull/428 earlier today, and @mih commented with 

> The should go into the extension template too, I think.

yarikoptic:
thank you!

PR #215, opened 2023-07-16
Add [extras] extras_require with datalad-metalad and add all those extras to [devel]

yarikoptic:
Development dependencies should be specified in setup.cfg as well, requirements*txt files should only reuse that setup.

Was missed during original PR code review I guess.


yarikoptic:
rebased, added changelog. if all good -- could be merged

Issue #219, opened 2023-09-11
No commandline documentation in this extension


adswa:
A comment in the DataLad channel by David Deepwell brought the following to my attention:

There is no commandline documentation for datalad-containers at all, and the one for the python API seems to be partially lacking:

![image](https://github.com/datalad/datalad-container/assets/29738718/5f0429bf-f4a9-4e15-bf2f-9ceea9a37aab)

And for the Python API, there are no docs at all for containers-run and containers-list. For containers-add and containers-remove the available docs are minimal at best:

![image](https://github.com/datalad/datalad-container/assets/29738718/6e4a2f75-41ce-49fa-b093-62045ded2b11)

adswa:
Building the docs locally however works fine, and results in usable docs:

![image](https://github.com/datalad/datalad-container/assets/29738718/68da0609-460d-41cc-bad3-e94198a5ae8e)

![image](https://github.com/datalad/datalad-container/assets/29738718/0f6df7c8-efda-4d75-bd25-631ea93b1049)

adswa:
fixed with #220

PR #220, opened 2023-09-11
Add readthedocs.yml config file from extension template

adswa:
The readthedocs build has been failing in recent builds, e.g. https://readthedocs.org/projects/datalad-container/builds/21818311/. This might be a reason behind #219. This PR adds the readthedocs.yml configuration from the datalad extension template, which is now required by readthedocs.


adswa:
The readthedocs build succeeded, and the preview looks good. I'm merging this in hopes it may also fix #219

PR #222, opened 2023-10-09
Update Travis badge

mih:


adswa (approved):
the diff is a bit hard to figure out, but I believe it fixes the travis bagde that is currently broken, and removes newlines and assures everything is in one line. I approve of that

PR #223, opened 2023-10-09
Link essential user documentation

mih:
This was previously needlessly too hard to find for people scanning the extension docs.


adswa (approved):
Thx, this works for me!

PR #227, opened 2023-10-09
Support `{python}` placeholder

### LINKED ISSUE #224: `containers-add` produces platform-specific configuration
```ini
[datalad "containers.debian"]
	image = .datalad\\environments\\debian\\image
	cmdexec = python.exe -m datalad_container.adapters.docker run {img} {cmd}
```

The python executable is not portable and the path separators also not.

### LINKED ISSUE #226: Support `{python}` placeholder for `cmdspec`
The way we handle the execution of the docker adaptor is suboptimal (see https://github.com/datalad/datalad-container/issues/224).

Even if we arrive at a more portable `python` exectuable, we still won't capture all variability across system. We need to get the *correct* python on the system that may rerun the record. It needs to be the one that has the adaptor installed.

it makes sense to me to start supporting a `{python}` placeholder that is replaced with `sys.exectuable` on execute. We would then know that this executable is the one that runs the datalad code that triggers the execution.

mih:
This placeholder is expanded on *execution* of a container, rather than on configuration/addition. This helps use the same Python installation that is also executing the datalad-container code.

Previously, the docker-support code would expand and then hardcode `sys.executable` on configuring a container. This led to non-portable configuration (e.g., hardcoded `python.exe` on windows), and would fail to pick up the correct python installation in any case where the `python` entrypoint would not point to the correct one.

Closes #226
Almost does #224 (missing the platform path issue)

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -65,7 +65,8 @@ def _guess_call_fmt(ds, name, url):
     elif url.startswith('shub://') or url.startswith('docker://'):
         return 'singularity exec {img} {cmd}'
     elif url.startswith('dhub://'):
-        return op.basename(sys.executable) + ' -m datalad_container.adapters.docker run {img} {cmd}'
+        # {python} is replaced with sys.executable on *execute*
+        return '{python} -m datalad_container.adapters.docker run {img} {cmd}'
 
 
 def _ensure_datalad_remote(repo):
@@ -143,6 +144,9 @@ class ContainersAdd(Interface):
             replaced with the desired command. Additional placeholders:
             '{img_dspath}' is relative path to the dataset containing the image,
             '{img_dirpath}' is the directory containing the '{img}'.
+            '{python}' expands to the path of the Python executable that is
+            running the respective DataLad session, for example a
+            'datalad containers-run' command.
             """,
             metavar="FORMAT",
             constraints=EnsureStr() | EnsureNone(),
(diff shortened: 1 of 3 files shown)


adswa:
The failures are due to dockerhubs rate limiting: 
```
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

datalad_container/containers_add.py
```diff
@@ -65,7 +65,8 @@ def _guess_call_fmt(ds, name, url):
     elif url.startswith('shub://') or url.startswith('docker://'):
         return 'singularity exec {img} {cmd}'
     elif url.startswith('dhub://'):
-        return op.basename(sys.executable) + ' -m datalad_container.adapters.docker run {img} {cmd}'
+        # {python} is replaced with sys.exectuable on execute
```
adswa, line 68:
```suggestion
        # {python} is replaced with sys.executable on execute
```

mih: Thanks, it is included in the next commit.

mih:
@adswa Yeah, the test setup is not so nice. Locally this works now -- also on windows.

adswa (approved):
looks sensible to me

yarikoptic (approved):

mih:
> any idea on how many tests would fail if windows is added as a test env?

When I execute the tests on windows (minus anything metalad-related https://github.com/datalad/datalad-metalad/issues/397 because I am working with PY3.12 right now), I see 8 failed, 13 passed.

A good chunk of the failures is simply https://github.com/datalad/datalad-container/issues/233, most of the rest is wrong path assumptions.

mih:
Rebased after merge of #235

mih:
Merging, given the approvals.

Issue #241, opened 2023-10-10
Missing `result_renderer='disabled'` for internal datalad command execution


mih:
This is the case for, at least, the `save` calls in `containers_add` and `containers_remove`. This should disable the internal rendering of results,
because otherwise a user cannot control the verbosity.

When this is addressed, it makes sense to also start using `return_type='generator'` to avoid the present list-item-accumulation of results of internal command calls.

Issue #240, opened 2023-10-10
Discontinue use of `datalad-remove` in `containers-remove`


mih:
The current implementation essentially uses `remove` via a chain of side-effect to do the equivalent of a `git rm` that avoids dropping a container image from the dataset.

This can be also achieved be achieved by removing the annex handle, and having the removal be (also) picked up by the final `datalad-save` that runs anyways.

Issue #239, opened 2023-10-10
Start using `register_config()`


mih:
https://github.com/datalad/datalad-container/blob/c505e52f02f9249e5f8edf478e5c45423d9dde18/datalad_container/containers_add.py#L235 is already asking for it.

This would make it possible to discontinue the approach in `definitions.py`

`register_config` was introduced with datalad v0.16

PR #242, opened 2023-10-10
Single-point for reading container configuration

### LINKED ISSUE #238: Implement standard utility to read container config
This is implemented individually in many places. This makes it needlessly hard to fix #224, where we decided to normalize on write and convert to platform specs on read. Given that all commands at least read the config, and standard helper will simplify things

mih:
This changeset established a new helper `get_container_configuration()` as the single method to read configuration on containers.

This closes #238 

This changeset is the precondition for implementing configuration item normalization effciciently -- for example as needed for resolving #224

A number of issues have been detected while working on this changeset, which are not addressed in here, but will be the topic of future work:

- #241 
- #240 
- #239

This changeset implies no behavior changes. The only user-facing change is an expanded documentation.

9b95b7862c4419fb9e3b69621a8d447a5d09ced0 is not strictly needed in this PR. It led to #241 and could be included in a future PR. However, I see no benefit it such a diff to accumulate dust.

### DIFF
--- a/datalad_container/utils.py
+++ b/datalad_container/utils.py
@@ -1,5 +1,7 @@
 """Collection of common utilities"""
 
+from __future__ import annotations
+
 from datalad.distribution.dataset import Dataset
 from datalad.support.external_versions import external_versions
 
@@ -15,6 +17,7 @@ def get_container_command():
 
 def get_container_configuration(
     ds: Dataset,
+    name: str | None = None,
 ) -> dict:
     """Report all container-related configuration in a dataset
 
@@ -26,6 +29,10 @@ def get_container_configuration(
     ----------
     ds: Dataset
       Dataset instance to report configuration on.
+    name: str, optional
+      If given, the reported configuration will be limited to the container
+      with this exact name. In this case, only a single ``dict`` is returned,
+      not nested dictionaries.
 
     Returns
     -------
@@ -34,6 +41,11 @@ def get_container_configuration(
       with their respective configuration items (with the
       ``datalad.containers.<container-name>.`` prefix removed from their
       keys).
+      If `name` is given, only a single ``dict`` with the configuration
+      items of the matching container is returned (i.e., there will be no
+      outer ``dict`` with container names as keys).
+      If not (matching) container configuration exists, and empty dictionary
+      is returned.
     """
     # all info is in the dataset config!
     var_prefix = 'datalad.containers.'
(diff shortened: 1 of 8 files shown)


adswa (approved):
this looks good to me!

datalad_container/utils.py
```diff
@@ -9,3 +15,62 @@ def get_container_command():
...
+      datalad.containers.<container-name>.<item-name>
+
+    Parameters
+    ----------
+    ds: Dataset
+      Dataset instance to report configuration on.
+    name: str, optional
+      If given, the reported configuration will be limited to the container
+      with this exact name. In this case, only a single ``dict`` is returned,
+      not nested dictionaries.
```
yarikoptic, line 35: could we make it more explicit, i.e. separate out a `get_containers_configuration(ds)` ?  internally `get_container_configuration(ds, name)` can just call that `get_containers_configuration(ds)`, I don't care.  But having singular semantic in the name but getting a dict for multiple containers will eventually confuse someone like me at some point because e.g. I thought without a name specified it would return the config for the "only" container (as e.g. `containers-run` behaves IIRC) and raise KeyError or smth like that if multiple are found -- so it becomes ambiguous.

yarikoptic (changes requested):
minor comment and 1 change requested to make API a bit clearer.

datalad_container/utils.py
mih: Just to clarify. Your response to the description in https://github.com/datalad/datalad-container/pull/242/commits/dbb2f914e43cf467858a2b7941ded686ae88908f (commit message) is: Let's nevertheless always read and post-process all configuration, make all filesystem checks for all containers, and then throw away all that information, but for one container.

Rather than that, you can also get that semantic cleanliness by having two near-copies of that function.

Just to be clear, I am not here to argue -- I just needs this merged, and I don't care in which way. I just do not want to guess what should be done and avoid future iterations on this code.

yarikoptic: I have missed that there is a `continue` shortcut, so indeed some code is skipped... even though complexity stays `O(N)`, there would indeed be not only a waste of cycles but also possibly errors on unrelated containers which should be avoidable. I guess a way could be to make current function `_private` and then expose it via two clean interfaces calling it out with or without a name.
I will leave it up to you whether to implement that or proceed with current version.

mih: This is a nice suggestion! I will change it to use a private helper and two clean interfaces.

Depending on your reviewing capabilities, I might do this in a separate follow-up PR, given that more PR are built on this, and they would all need an update.

yarikoptic: Then let's merge

PR #243, opened 2023-10-10
Unconditionally store image path in POSIX notation in configuration

### LINKED ISSUE #224: `containers-add` produces platform-specific configuration

mih:
[This PR sits on #242, only the last commit is unique and relevant here]

This establishes a fixed format for the configuration. Previously,
the path was stored in platform conventions without any indication
of the nature of that platform. This turned platform detection into
guesswork. Configuration created on windows was not functional on other platforms.

Some of that guesswork is now implemented in a path normalization
helper (`_normalize_image_path()`). It is executed on-read, in order
to keep older dataset configuration functional for the cases

- we can tell the platform conventions, because the image is placed
under `.datalad\\environments\\` (the default)
- we can locate a matching element on the file system

Notably, this guesswork does not cover the case of an image that
was configured on windows, and is hosted in a currently uninstalled
subdataset. This seems rare enough of a case to not deserve the implied complexity of supporting it in a backward compatible way.

Closes #224

TODO:

- [x] run this on windows and confirm that it works (stores POSIX, and operates on platform paths)
- [x] adjust tests to simulate an old setup with windows paths

The missing lines in the coverage are

image

I am not planning on adding a test for this case (a path that is verified to be in windows convention by finding a matching filesystem item). If this is unacceptable, I am OK with removing support for this entirely.

### DIFF
--- a/datalad_container/utils.py
+++ b/datalad_container/utils.py
@@ -2,6 +2,15 @@
 
 from __future__ import annotations
 
+# the pathlib equivalent is only available in PY3.12
+from os.path import lexists
+
+from pathlib import (
+    PurePath,
+    PurePosixPath,
+    PureWindowsPath,
+)
+
 from datalad.distribution.dataset import Dataset
 from datalad.support.external_versions import external_versions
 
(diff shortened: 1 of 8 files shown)


datalad_container/utils.py
```diff
@@ -9,3 +24,128 @@ def get_container_command():
...
+
+        cinfo = containers.get(cname, {})
+        cinfo[ccfgname] = value
+
+        containers[cname] = cinfo
+
+    return containers if name is None else containers.get(name, {})
+
+
+def _normalize_image_path(path: str, ds: Dataset) -> str:
```
yarikoptic, line 95: I guess we aren't running mypy here...

```suggestion
def _normalize_image_path(path: str, ds: Dataset) -> PurePath:
```

datalad_container/utils.py
```diff
@@ -9,3 +24,128 @@ def get_container_command():
...
+    """Helper to standardize container image path handling
+
+    Previously, container configuration would contain platform-paths
+    for container image location (e.g., windows paths when added on
+    windows, POSIX paths elsewhere). This made cross-platform reuse
+    impossible out-of-the box, but it also means that such dataset
+    are out there in unknown numbers.
+
+    This helper inspects an image path READ FROM CONFIG(!) and ensures
+    that is matches platform conventions (because all other arguments)
```
yarikoptic, line 105:
```suggestion
    that it matches platform conventions (because all other arguments)
```

datalad_container/utils.py
```diff
@@ -9,3 +24,128 @@ def get_container_command():
...
+    also come in platform conventions. This enables standardizing
+    the storage conventions to be POSIX-only (for the future).
+
+    Parameters
+    ----------
+    path: str
+      A str-path, as read from the configuration, matching its conventions
+      (relative path, pointing to a container image relative to the
+      dataset's root.
```
yarikoptic, line 114:
```suggestion
      dataset's root).
```

datalad_container/utils.py
```diff
@@ -9,3 +24,128 @@ def get_container_command():
...
+    """
+    # we only need to act differently, when an incoming path is
+    # windows. This is not possible to say with 100% confidence,
+    # because a POSIX path can also contain a backslash. We support
+    # a few standard cases where we CAN tell
+    pathobj = None
+    if '\\' not in path:
+        # no windows pathsep, no problem
+        pathobj = PurePosixPath(path)
+    elif path.startswith('.datalad\\environments\\'):
```
yarikoptic, line 132: are you sure you don't need a
```suggestion
    elif path.startswith(r'.datalad\\environments\\'):
```
?  IIRC you showed that it really stores `\\` in the path in the config, or it converts them to a single `\` upon loading?  (in the [test above](https://github.com/datalad/datalad-container/pull/243/files#diff-c9a6a0cde74f0c35ded765fc396d537f9485fe47645c4e00698ab951138042d6R408) you also set to a single `\` since lacking `r""` so I wondered to double check if that corresponds to actual target to fix)

datalad_container/utils.py
yarikoptic: the rest is not English for me either, but I am not quite certain what was the meaning intended to be delivered, so leaving asis

yarikoptic (approved):
conditionally on that non-r""-ed use of `\\` in strings.  IIRC it was literally double `\\` in the file.

datalad_container/utils.py
mih: Oh man, wtf!? you are absolutely correct!! My brain did an internal `print()` all these days. I just went back and confirmed again that the report in https://github.com/datalad/datalad-container/issues/224 is correct and we do have `\\` in the config files generated on windows, not `\`

:facepalm:

mih:
Thanks for the reviews! Very helpful!

PR #246, opened 2023-10-10
Docker adaptor can retrieve an image from the local docker service

### LINKED ISSUE #199: docker: do not get layers in present in local docker service

mih:
Closes: #199

@bpinsard would be great if you could give this a spin!

Demo

```
❯ datalad create busydemo
❯ cd busydemo
❯ datalad containers-add -u dhub://busybox:latest busy
❯ datalad drop .datalad/environments --reckless availability
❯ git annex info | grep 'local annex'
local annex keys: 0
local annex size: 0 bytes
❯ cat .datalad/config
[datalad "dataset"]
        id = b7adee52-a65a-43fc-a85b-c0d5e2d5b67c
[datalad "containers.busy"]
        image = .datalad/environments/busy/image
        cmdexec = {python} -m datalad_container.adapters.docker run {img} {cmd}
❯ datalad containers-run -n busy uname
[INFO   ] Saved busybox:latest to /tmp/busydemo/.datalad/environments/busy/image/tmpzrdd7mj2
[INFO   ] Making sure inputs are available (this may take some time)
[INFO   ] == Command start (output follows) =====
Linux
[INFO   ] == Command exit (modification check follows) =====
run(ok): /tmp/busydemo (dataset) [/home/mih/env/datalad-dev/bin/python -m ...]
```

### DIFF
--- a/datalad_container/adapters/docker.py
+++ b/datalad_container/adapters/docker.py
@@ -13,6 +13,7 @@ import hashlib
 import json
 import os
 import os.path as op
+from pathlib import Path
 import subprocess as sp
 import sys
 import tarfile
@@ -88,6 +89,14 @@ def _list_images():
     return out.decode().splitlines()
 
 
+def _get_repotag_from_image_sha256(sha):
+    out = sp.check_output(
+        ['docker', 'image', 'inspect', '--format',
+         '{{range $v := .RepoTags}}{{$v}} {{end}}',
+         sha])
+    return out.decode().splitlines()[0].strip()
+
+
 def get_image(path, repo_tag=None, config=None):
     """Return the image ID of the image extracted at `path`.
     """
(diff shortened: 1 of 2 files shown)


bpinsard:
Works for me, it seems to retrieve the layers correctly.

I currently no longer have a specific usecase for the docker adapter, but that is a great performance enhancement.

yarikoptic:
sorry, not grasping atm the whole machinery... if code works for you -- go ahead with the merge.

mih:
Thanks for the feedback! I am not sure whether I will have the chance to try the practicality of this implementation sufficiently. The next week will tell me that. For now, I will convert this to a draft PR.

If I get this chance to try this out sufficiently I will finish it out and repropose.

If this message is read after two weeks from now, and the PR is still in draft. Please close without waiting for me. Thanks!

PR #247, opened 2023-10-10
Add protection against invalid image path declarations for `add`

### LINKED ISSUE #150: containers-add should throw error message when referenced image is outside of dataset
Running `datalad containers-add fMRIprep -i '/data/project/singularity/fmriprep-20.2.0.simg' -d '/data/project/hcp_movies/hcp-openaccess/'` did not result in an error even though data/project/singularity/fmriprep-20.2.0.simg does not exist in the dataset.

mih:
This is a bit painful, because datalad-next parameter validation is not available, but worth adding given the issue.

Closes #150

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -199,6 +199,16 @@ class ContainersAdd(Interface):
                              purpose='add container')
         runner = WitlessRunner()
 
+        if image is not None:
+            if op.isabs(image):
+                raise ValueError(
+                    f'image must be a relative path, got {image!r}')
+            if ds.pathobj.resolve() \
+                    not in (ds.pathobj / image).resolve().parents:
+                raise ValueError(
+                    'image must be a relative path pointing inside'
+                    f'the dataset, got {image!r}')
+
         # prevent madness in the config file
         if not re.match(r'^[0-9a-zA-Z-]+$', name):
             raise ValueError(
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -206,7 +206,7 @@ class ContainersAdd(Interface):
             if ds.pathobj.resolve() \
                     not in (ds.pathobj / image).resolve().parents:
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


mih:
Test failures reveal that internal use of the parameter is also inconsistent with its documented semantics.

datalad_container/containers_add.py
```diff
@@ -199,6 +199,16 @@ def __call__(name, url=None, dataset=None, call_fmt=None, image=None,
...
         runner = WitlessRunner()
 
+        if image is not None:
+            if op.isabs(image):
+                raise ValueError(
+                    f'image must be a relative path, got {image!r}')
+            if ds.pathobj.resolve() \
+                    not in (ds.pathobj / image).resolve().parents:
+                raise ValueError(
+                    'image must be a relative path pointing inside'
```
yarikoptic, line 209:
```suggestion
                    'image must be a relative path pointing inside '
```

datalad_container/containers_add.py
```diff
@@ -199,6 +199,16 @@ def __call__(name, url=None, dataset=None, call_fmt=None, image=None,
                              purpose='add container')
         runner = WitlessRunner()
 
+        if image is not None:
+            if op.isabs(image):
```
yarikoptic, line 203: Docstring does say relative so I guess it is good. I just wanted to raise a comment that on general we are ok with providing absolute paths everywhere and I think we (or git) normalize them into relative if needed... so it would've been more consistent to just normalize absolute path to correct relative here

yarikoptic:
Appveyor shows now failing test_run_subdataset_install because path isn't absolute... I guess somewhat reflects on my comment above

mih:
After looking at this again, I decided to close this PR. There would be a significant effort resulting in the serial parameter validation that the `ValidatedInterface` and the associated tooling in `datalad-next` is trying to overcome.

I think this is better approached once that or an equivalent functionality is available to this extension.

PR #250, opened 2023-10-12
Stop substitution `{python}` at `containers-run` runtime

### LINKED ISSUE #249: `{python}` placeholder implementation non-portable
I added the support for the placeholder in `containers-run` with #227. This means it will get resolved before `run` can see it, and it will enter the run record in resolved form.

This is bad, because this resolved form is as non-portable as the previous setup, if not more.

This change must be amended, the resolution must be performed in `run`, and here we need to use the placeholder in its escaped form: `{{python}}`.

A viable approach might be to define `datalad.run.substitutions.python=sys.executable`

mih:
Stop substitution `{python}` at `containers-run` runtime
    
This causes worse portability issues than the previous approach of storing `python(.exe)`.
    
Instead pass the placeholder on to `datalad run`.
    
This would now error with
    
```
command has an unrecognized placeholder: 'python'
```
    
and requires a further intervention. Three options:

- have datalad-core define `datalad.run.substitutions.python=sys.executable`
- wait for https://github.com/datalad/datalad-container/pull/244 and use `register_config()` to have datalad-container define it
- have datalad-container define it at `containers-run` runtime by patching the configuration for the execution time (would have `rerun` fail still)
- add the configuration item to the committed dataset config
    
Closes #249

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -69,8 +69,9 @@ def _guess_call_fmt(ds, name, url):
     elif url.startswith('shub://') or url.startswith('docker://'):
         return 'singularity exec {img} {cmd}'
     elif url.startswith('dhub://'):
-        # {python} is replaced with sys.executable on *execute*
-        return '{python} -m datalad_container.adapters.docker run {img} {cmd}'
+        # {{python}} is eventually substituted with something like
+        # `sys.executable` on *execute* by datalad-run
+        return '{{python}} -m datalad_container.adapters.docker run {img} {cmd}'
 
 
 def _ensure_datalad_remote(repo):
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -102,7 +102,82 @@ class ContainersAdd(Interface):
     # first docstring line is used a short description in the cmdline help
     # the rest is put in the verbose help and manpage
     """Add a container to a dataset
+
+    Adding a container is primarily placing container-related configuration in
+    the committed dataset configuration at ``.datalad/config``. At minimum,
+    only two bits of information are required:
+
+    - location of the container image
+    - specification of how the images is to be executed
+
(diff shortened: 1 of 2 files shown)


yarikoptic:
#244 was merged, should we proceed that way?

mih:
Sadly, this does not seem to be working. It looks like `run` is not considering configuration defaults, but only explicit configuration:

```py
>>> import datalad
>>> import datalad_container
>>> datalad.cfg.get('datalad.run.substitutions.python') is None
True
>>> datalad.cfg.obtain('datalad.run.substitutions.python')
'/home/mih/env/datalad-dev/bin/python'
```

mih:
Here is a patch for `run()` that would make this work.

```diff
diff --git a/datalad/core/local/run.py b/datalad/core/local/run.py
index a2ca86772..b588c4a34 100644
--- a/datalad/core/local/run.py
+++ b/datalad/core/local/run.py
@@ -38,6 +38,7 @@ from datalad.interface.base import (
     build_doc,
     eval_results,
 )
+from datalad.interface.common_cfg import definitions as cfg_defs
 from datalad.interface.common_opts import (
     jobs_opt,
     save_message_opt,
@@ -623,7 +624,12 @@ def format_command(dset, command, **kwds):
     command = normalize_command(command)
     sfmt = SequenceFormatter()
 
-    for k, v in dset.config.items("datalad.run.substitutions"):
+    for k in set(cfg_defs.keys()).union(dset.config.keys()):
+        v = dset.config.get(
+            k,
+            # pull a default from the config definitions
+            # if we have no value, but a key
+            cfg_defs.get(k, {}).get('default', None))
         sub_key = k.replace("datalad.run.substitutions.", "")
         if sub_key not in kwds:
             kwds[sub_key] = v
```

I think this would be a worthwhile addition.

mih:
I posted a corresponding PR at https://github.com/datalad/datalad/pull/7509

Issue #254, opened 2024-01-16
release fails since build dependencies aren't installed while asking for the version


yarikoptic:
we are using

```
      - name: Prepare release
        uses: datalad/release-action/release@master
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          pypi-token: ${{ secrets.PYPI_TOKEN }}
          pre-tag: |
            pkg="$(python setup.py --name)"
            printf '__version__ = "%s"\n' "$new_version" > "$pkg"/version.py
            git commit -m "Update __version__ to $new_version" "$pkg"/version.py
```

but now that step fails with 

```
Run pkg="$(python setup.py --name)"
Traceback (most recent call last):
  File "/home/runner/work/datalad-container/datalad-container/setup.py", line 3, in <module>
    from setuptools import setup
ModuleNotFoundError: No module named 'setuptools'
```

@jwodder - do you have an immediate recipe for the remedy?  

should we just  `pip install .` in `pre-tag` before running that pkg= ? but that would install all unnecessary stuff... we could just `pip install setuptools` but that feels ad-hoc

jwodder:
@yarikoptic 

* Why query `pkg` in this way instead of just writing "`datalad_container`"?
* As part of the move towards `pyproject.toml`-based builds, [invoking `python setup.py <anything>` is now deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html), and `setuptools` is no longer bundled with Python (hence the error you're seeing).

> should we just `pip install .` in `pre-tag` before running that pkg= ?

I fail to see how doing that would make `python setup.py --name` succeed.

yarikoptic:
> @yarikoptic
> 
> * Why query `pkg` in this way instead of just writing "`datalad_container`"?

I guess that in eeea9db4feb0b2f1c916ddf3a35406eae4207a2f you tried to use some generic pattern so we could use s.across various packages.  but yeah -- could be hardcoded I guess.

> * As part of the move towards `pyproject.toml`-based builds, [invoking `python setup.py <anything>` is now deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html), and `setuptools` is no longer bundled with Python (hence the error you're seeing).

I see. good. I guess we could follow up my dirty fixes with hardcoding the name there and thus simplifying it all and that would close this issue

PR #255, opened 2024-01-16
Workaround for now missing setuptools which are used in setup.py

yarikoptic:
Ref: https://github.com/datalad/datalad-container/issues/254

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           pypi-token: ${{ secrets.PYPI_TOKEN }}
           pre-tag: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -27,6 +27,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           pypi-token: ${{ secrets.PYPI_TOKEN }}
           pre-tag: |
+            pip install setuptools  # needed for setup.py --version
```
jwodder, line 30: @yarikoptic This is for `setup.py --name`, not `setup.py --version` (which you're not doing).

yarikoptic: dang... right... damn suggestions in pycharm... will all be gone with hardcoding...

PR #257, opened 2024-01-17
BF: move check for metalad before import of meta_extract

yarikoptic:
Otherwise we try to import meta_extract and it fails.  Failed the build of datalad-container for debian

I will shortly bundle also run of isort on that file to ensure that pragmas are effective

### DIFF
--- a/datalad_container/extractors/tests/test_metalad_container.py
+++ b/datalad_container/extractors/tests/test_metalad_container.py
@@ -1,20 +1,24 @@
 import os.path as op
-import pytest
 import subprocess
 import sys
 from pathlib import Path
 from shutil import which
 from unittest.mock import patch
 
+import pytest
+from datalad.support.external_versions import (
+    UnknownVersion,
+    external_versions,
+)
 # Early detection before we try to import meta_extract
-from datalad.tests.utils_pytest import SkipTest  # noqa: E402, isort:skip
-from datalad.support.external_versions import external_versions, UnknownVersion  # noqa: E402, isort:skip
+from datalad.tests.utils_pytest import SkipTest
+
 if not external_versions["datalad_metalad"]:
     raise SkipTest("skipping metalad tests")
 
 from datalad.api import (
-    clone,
     Dataset,
+    clone,
     meta_extract,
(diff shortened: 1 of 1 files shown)


datalad_container/extractors/tests/test_metalad_container.py
```diff
@@ -1,22 +1,31 @@
...
+# Early detection before we try to import meta_extract
+from datalad.tests.utils_pytest import SkipTest
+
+if not external_versions["datalad_metalad"]:
+    raise SkipTest("skipping metalad tests")
+
 from datalad.api import (
-    clone,
     Dataset,
+    clone,
```
asmacdo, line 21: Is there a convention to import classes first? I've always just done strictly alphabetical

asmacdo (approved):
one nitpick/question LGTM

datalad_container/extractors/tests/test_metalad_container.py
yarikoptic: "not sure".  Please find an answer at https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html somewhere or may be @jwodder could clarify?

yarikoptic:
releasing so we simplify update of packages for debian and conda-forge both of which need this now

datalad_container/extractors/tests/test_metalad_container.py
asmacdo: isort does sort by case

https://pycqa.github.io/isort/docs/configuration/options.html

```
Order imports by type, which is determined by case, in addition to alphabetically.

NOTE: type here refers to the implied type from the import name capitalization. isort does not do type introspection for the imports. These "types" are simply: CONSTANT_VARIABLE, CamelCaseClass, variable_or_function. If your project follows PEP8 or a related coding standard and has many imports this is a good default, otherwise you likely will want to turn it off. From the CLI the --dont-order-by-type option will turn this off.

Type: Bool
Default: True
Config default: true
Python & Config File Name: order_by_type
CLI Flags:

--ot

--order-by-type
```

PR #259, opened 2024-01-17
Hardcode the name of the package in release action to avoid needing setuptools and invocation of setup.py

### LINKED ISSUE #254: release fails since build dependencies aren't installed while asking for the version

yarikoptic:
Per https://github.com/datalad/datalad-container/issues/254#issuecomment-1895808006

Closes #254

no rush for this IMHO so not marking for release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,9 +27,8 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           pypi-token: ${{ secrets.PYPI_TOKEN }}
           pre-tag: |
-            pip install setuptools  # needed for setup.py --version
-            pkg="$(python setup.py --name)"
-            printf '__version__ = "%s"\n' "$new_version" > "$pkg"/version.py
-            git commit -m "Update __version__ to $new_version" "$pkg"/version.py
+            version_file=datalad_container/version.py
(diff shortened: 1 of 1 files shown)


jwodder (approved):

yarikoptic:
dang, how could I forget to add `release` label? uff...

PR #260, opened 2024-01-17
Run isort across entire codebase

yarikoptic:
- for consistency since we already have `isort` configuration
- to trigger a release ;)

I hope there would be no side-effects on the order of imports -- all trust goes into tests!


yarikoptic:
I don't see some obvious issues, CI is happy, let's see what happens

PR #262, opened 2024-02-01
Improving documentation for --url parameter

mhaid:
Adding documentation regarding the default behaviour of "cli_run" in adapter "docker.py", which is called when --url should fetch from docker-hub.

Maybe it is also helpful for the user to mention the use of the "docker adapter" in the docs.

### DIFF
--- a/datalad_container/containers_add.py
+++ b/datalad_container/containers_add.py
@@ -138,7 +138,8 @@ class ContainersAdd(Interface):
             the rest of the URL will be interpreted as the argument to
             'docker pull', the image will be saved to a location
             specified by `name`, and the call format will be auto-configured
-            to run docker, unless overwritten.""",
+            to run docker, unless overwritten. The auto-configured call to docker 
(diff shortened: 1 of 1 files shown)


datalad_container/containers_add.py
```diff
@@ -138,7 +138,8 @@ class ContainersAdd(Interface):
             the rest of the URL will be interpreted as the argument to
             'docker pull', the image will be saved to a location
             specified by `name`, and the call format will be auto-configured
-            to run docker, unless overwritten.""",
+            to run docker, unless overwritten. The auto-configured call to docker 
+            run mounts the cwd to '/tmp' and sets the working directory to '/tmp'.""",
```
yarikoptic, line 142: Thanks! really that is what it does? yikes -- something we might want to improve to avoid committing all the junk -- but it indeed might be good for "complete provenance collection" ;)  meanwhile since CWD is an acronym

```suggestion
            run mounts the CWD to '/tmp' and sets the working directory to '/tmp'.""",
```

adswa:
Any reason not to merge this, @yarikoptic?

Issue #263, opened 2024-02-02
Singularity installation fails with missing dependency


adswa:
The Appveyor run in #262 failed during setup with

```
Preparing to unpack /tmp/singularity-ce.deb ...
Unpacking singularity-ce (4.1.1-focal) ...
dpkg: dependency problems prevent configuration of singularity-ce:
 singularity-ce depends on fuse2fs; however:
  Package fuse2fs is not installed.
dpkg: error processing package singularity-ce (--install):
 dependency problems - leaving unconfigured
Processing triggers for man-db (2.9.1-1) ...
Errors were encountered while processing:
 singularity-ce
Command exited with code 1
Running "on_finish" scripts
```

Adding this dependency should resolve the issue.

PR #264, opened 2024-02-02
CI/tools: Add fuse2fs dependency for singularity installation

adswa:
Should provide a fix for #263


adswa:
The singularity installation succeeds now.
There is a test failure, but I don't think it needs addressing in this PR:

```
============================================================================================================
FAILED ../adapters/tests/test_docker.py::test_load_multi_image - datalad.runner.exception.CommandError: CommandError: 'docker pull alpine:3.15' failed with exitcode 1
======================================================================================= 1 failed, 28 passed, 
```

yarikoptic:
I restarted that appveyor run -- apparently was some intermittent failure, so we are all good.  I do not think we need a release for it but I could be proven to be wrong ;)

PR #268, opened 2024-09-16
MNT: Account for a number of deprecations in core

adswa:
this reduces the number of deprecation warnings when running the test suite


datalad_container/tests/test_containers.py
```diff
@@ -107,7 +107,7 @@ def test_container_files(ds_path=None, local_file=None, url=None):
     # non-default location:
     ds.config.add("datalad.containers.location",
                   value=op.join(".datalad", "test-environments"),
-                  where='dataset')
+                  scope='branch')
```
yarikoptic, line 110: currently we have `datalad > 0.18.0` and scope came in 0.16.0 . the other changes (drop and install) I think even before then so we should be good. Thank you @adswa !

PR #271, opened 2025-02-05
Add pre-commit config and "apply" it - no trailing spaces, sorted imports

yarikoptic:
Also removed duplicate `.codespellrc` since now we specify in pyproject.toml

Inspired by seeing all trailing spaces deleted by @mslw in 

- #270


yarikoptic:
appveyor -- seems overall in a bad shape

```
[ "x$PY" != x ] && . ${HOME}/venv${PY}/bin/activate || python3 -m venv ${HOME}/dlvenv && . ${HOME}/dlvenv/bin/activate; ln -s "$VIRTUAL_ENV" "${HOME}/VENV"
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.
    apt install python3.8-venv
You may need to use sudo with that command.  After installing the python3-venv
package, recreate your virtual environment.
Failing command: ['/home/appveyor/dlvenv/bin/python3', '-Im', 'ensurepip', '--upgrade', '--default-pip']
ln: failed to create symbolic link '/home/appveyor/VENV' -> '': No such file or directory
Command exited with code 1
Running "on_finish" scripts
while [ -f ~/BLOCK ]; do sleep 5; done
Build failed
```

should be fixed separately.  I do not expect breakages but apparently we do not have any unittesting going on without this appveyor thing... heh, will look into fixing here, but in the longer run I think we need to centralize such testing, filed

- https://github.com/datalad/datalad-extensions/issues/132

PR #273, opened 2025-05-17
BF: use setuptools.errors.OptionError instead of now removed import of distutils.DistutilsOptionError

yarikoptic:
setuptools.errors were introduced in 59.0.0 release.

See https://github.com/pypa/setuptools/issues/5000 for more rationale
etc.


yarikoptic:
ok, I see neither interest nor objections, so will proceed to make CIs stop spamming me daily.

